PR #8, opened 2022-06-06
Add initial pipeline documentation + add generate binaries page

IgorTodorovskiIBM:
* Adds initial pipeline documentation
* Also adds getbinaries.py to generate the Latest.md file.  This will run as part of Jenkins and generate on demand.


docs/Guides/Pipeline.md
```diff
@@ -0,0 +1,41 @@
...
+The z/OS Open Tools CI/CD pipeline leverages z/OS Virtual Server Instances (zVSI) to be able to build, test, and deploy ported packages to Github Releases.
+
+There is currently one zVSI Jenkins node set up under the ip address `128.168.136.67`.
+
+### zVSI Directory Structure
+* TBD
+
+### z/OS Open Tools Jenkins CI/CD Pipeline
+
+Access the Jenkins CI/CD Pipeline at https://128.168.139.253:8443.  To be able to view the Jenkins CI/CD jobs, you must first log-in with your w3id.
```
MikeFultonDev, line 16: W3 id? Why does it have to be w3?

MikeFultonDev (commented):
See my comment about w3. Otherwise looks good

docs/Guides/Pipeline.md
IgorTodorovskiIBM: We can change it, but since we haven't exposed the zVSI, I authenticate against github.ibm.com which uses w3id.

PR #25, opened 2022-08-16
Update documentation

IgorTodorovskiIBM:
* Added a started page
* Made z/OS more consistent
* Fixed the "edit this page" link
* Renamed sidebar links

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,9 +1,13 @@
 ## Welcome to z/OS Open Tools
 
-Welcome to the z/OS Open Tools documentation. 
+Welcome to the **z/OS Open Tools documentation**. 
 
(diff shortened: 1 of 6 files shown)


docs/README.md
```diff
@@ -1,11 +1,7 @@
-# ZOSOpenTools docs
+## Welcome to z/OS Open Tools
 
-To make changes to the docs, please:
```
MikeFultonDev, line 3: i agree this part shouldn't be on the front page, but did it completely disappear?

IgorTodorovskiIBM: Thanks, I added a reference to it in the latest commit:

Screen Shot 2022-08-16 at 12 03 15 PM

PR #38, opened 2022-09-02
added a note about creating patches from directories in .gitignore

ejratl:
The instructions in the porting guide for creating patches only work if the open source tool has been cloned via git. For the example and for some of the existing ports, the open source tool code is sourced from a tarball. In that case, the patches will have to be manually created with diff. I tried to highlight this in this update, but the version of diff on the zOS instance that I am using is missing some of my favorite diff options.

I am not sure whether git apply will be able to use the patches created manually using diff. If git can't use the patches, then the document should be amended to say that patches can only be created when the tool source is cloned via git, and not when the tool source comes from a tarball.

I also needed the environment sourcing command before I could get anything to work, so I thought it would be a useful addition to this document. I wasn't really sure where to put it.

Also some whitespace cleanup.

### DIFF
--- a/docs/Guides/Porting.md
+++ b/docs/Guides/Porting.md
@@ -8,7 +8,7 @@ Before you begin to build code, ensure that you have access to a z/OS UNIX envir
 
 ### Set up your environment
 
-This is the most cumbersome step, but is only required to be done once. 
+This is the most cumbersome step, but is only required to be done once.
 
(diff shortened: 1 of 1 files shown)


IgorTodorovskiIBM (approved):
LGTM

MikeFultonDev (approved):

MikeFultonDev:
Thanks for the updates!

PR #41, opened 2022-09-09
add example for open() and tagging

MikeFultonDev:


IgorTodorovskiIBM (approved):
LGTM

ejratl (approved):
This looks good to me. As a person without a zOS background, the two things that I would like to have further explained are 1)  the 819 == ASCII relationship (where do I look this up) and 2) (related but orthogonal to the goal here) a quick hint about whether files can be tagged from the command line, once they have been created without the tagging (this will help me decide how to more forward once I have fixed the issue).

Also, Markdown linters really want blank lines before and after code segments.

ejratl (approved):
Exceptionally helpful for people like me with an open source rather than an IBM z background. Thanks for adding this additional information, since I will have to use it for the port that I am working on.

PR #43, opened 2022-09-15
expand on FSUM problem

MikeFultonDev:
improved the docs to explain how to eliminate the annoying messages about signal 13.
A side effect is that we should likely see if we should get changes to autoconf etc made to propagate this fix or whether we should just push on LE to support signal 13 properly.


IgorTodorovskiIBM (approved):
Useful! LGTM

PR #69, opened 2022-11-04
mention rocket curl for consumers

covener:
Both GH release one-liners and `zopen download` depend on pre-existing curl.

`zopen download` will tolerate older curl, but the one liners on the release pages fail with 7.42 (circa 2017). It is some pipe translation issue that results in:
>  pax: checksum error on tape (got f5bd, expected 0)


IgorTodorovskiIBM (approved):
LTGM

MikeFultonDev (approved):

PR #70, opened 2022-11-06
Getting started

MikeFultonDev:
This pull request will bring in a new utility called zopen-setup.
This utility is available in the 'release' of meta with instructions on how to use.

I have done a 'sample run' of it but would love it if the reviewers could as well to ensure it works ok for them too.


tools/src/zopen_boot_uri.h
```diff
@@ -0,0 +1,7 @@
+#ifndef ZOPEN_BOOT_URI
+  #include <stdlib.h>
+  #define ZOPEN_BOOT_URI_PREFIX "ZOSOpenTools"
+  #define ZOPEN_BOOT_URI_SUFFIX "releases/download/boot"
+  #define ZOPEN_BOOT_URI_TYPE "zos.pax.Z"
+  #define ZOPEN_BOOT_PKG { "gzip", "curl", "make", "utils", NULL }
```
IgorTodorovskiIBM, line 6: can you also add tar here

MikeFultonDev: done

PR #72, opened 2022-11-08
Update Pre-req.md

IgorTodorovskiIBM:


docs/Guides/Pre-req.md
```diff
@@ -22,35 +22,35 @@ It is recommended that you add the above environment variables to your `.profile
...
 ```bash
 zopen download --list
 ```
 
-To download and install a specific package, you can specify the -r option as follows:
+To download and install specific packages, specify them as a comma delimited list as follows:
 
 ```bash
-zopen download -r makeport
+zopen download make,cur,gzip
```
MikeFultonDev, line 53: should be curl not cur

IgorTodorovskiIBM, line 53:
```suggestion
zopen download make,curl,gzip
```

MikeFultonDev (approved):

PR #90, opened 2022-11-30
Remove colons from headings

ijmitch:
They look out of place in the left hand navigation and I see no reason for them.


IgorTodorovskiIBM (approved):
LGTM

PR #95, opened 2022-12-16
Better error message when destination is read-only

AnthonyGiorgio:
I tried installing to a directory that I didn't have write permissions for, and the error message wasn't very obvious.  This PR adds a nice and verbose reason as to why the failure occurred.


MikeFultonDev (approved):
LGTM

PR #96, opened 2022-12-16
Adjust directory permissions

AnthonyGiorgio:
The permissions on the subdirectories created under the zopen base directory are too restrictive by default.  They are 0740, which isn't useful for anyone except the owner.  This patch adjusts them to be 0755, so all users across the system can access the tools.

This also adjusts the `.gitignore` file to ignore vim backup and swap files.


IgorTodorovskiIBM (approved):
LGTM

PR #97, opened 2022-12-16
Enable compiler informational messages, and fix the results

AnthonyGiorgio:
I noticed a call to `fprintf` that had too many arguments for the number of formatters in the string, so I fixed that.  Then I wondered why it wasn't called out by the compiler, and I noticed that the informational messages weren't enabled.  I fixed that, and then got a whole bunch of warnings across the source tree.  I fixed all **that**, and finally got a clean compile.

### DIFF
--- a/tools/src/Makefile
+++ b/tools/src/Makefile
@@ -3,7 +3,8 @@
 #
 CC=xlc
 LD=xlc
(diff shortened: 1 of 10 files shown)


MikeFultonDev (approved):
Nice improvements. Thank you

PR #98, opened 2022-12-16
Support Github oauth token for downloads

AnthonyGiorgio:
This PR adds support for the Github oath token during the download step.  This supports the same `ZOPEN_GIT_OAUTH_TOKEN` environment variable that the `zopen` tool does.

### DIFF
--- a/tools/src/download.c
+++ b/tools/src/download.c
@@ -2716,14 +2716,14 @@ int setRequestHeaders( HWTH_RETURNCODE_TYPE *rcPtr,
  *
  * Returns: Address of heap-allocated ptr array
  ****************************************************************/
-char **getRequestHeaders() {
+char **getRequestHeaders(void) {
  char buf[80];
  char **headers = NULL;
- int NUM_HEADERS = 2;
+ int num_headers = 1;
 
  char *token = getenv("ZOPEN_GIT_OAUTH_TOKEN");
- if (token == NULL) {
-	token = "";
+ if (token != NULL) {
+	num_headers++;
  }
 
 /***************************************************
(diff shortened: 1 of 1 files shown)


tools/src/download.c
```diff
@@ -2719,20 +2719,30 @@ int setRequestHeaders( HWTH_RETURNCODE_TYPE *rcPtr,
...
+ if (token == NULL) {
+	token = "";
+ }
 
 /***************************************************
  * Allocate an array of pointers to string, and
  * build a string for each of the headers.  Use a
  * NULL last array element to indicate end of array.
  ***************************************************/
  headers = (char **)calloc( 1+NUM_HEADERS, sizeof(char *) );
```
MikeFultonDev, line 2734: malloc is good enough right? can you error out if the memory request fails (yes - i know i should've done that but since you're playing here maybe it's a good time to enhance it?

tools/src/download.c
```diff
@@ -2719,20 +2719,30 @@ int setRequestHeaders( HWTH_RETURNCODE_TYPE *rcPtr,
...
- sprintf( buf,
+ snprintf( buf, sizeof buf,
           "%s:%s",
           "User-Agent",
           "toolkit" );
- headers[0] = (char *)strdup(buf);
- headers[1] = NULL;
+ headers[0] = strdup(buf);
+
+ snprintf(buf, sizeof buf,"%s:%s %s",
```
MikeFultonDev, line 2741: is this a valid header if token = "" ?

AnthonyGiorgio: It worked when I tested it, but I can rework the code to make it cleaner.  My excuse was that I was in a hurry to get it checked in.

tools/src/download.c
AnthonyGiorgio: I've added a check for the `calloc()` call.

MikeFultonDev (approved):
Lgtm

PR #102, opened 2022-12-27
Add pre-req project links and license info

ijmitch:


README.md
```diff
@@ -33,20 +32,24 @@ First, you need to have some tools installed on your system:
 
 ### System Pre-reqs:
 
- - gnu make 4.1: Download unsupported binary from Rocket
- - xlclang 2.4.1: Download unsupported binary from IBM
- - git: Download unsupported binary from Rocket
- - curl: Download unsupported binary from Rocket
- - gunzip: Download unsupported binary from Rocket
+ - [gnu make 4.1](https://www.gnu.org/software/make/)[^gpl] - Download unsupported binary from Rocket
```
MikeFultonDev, line 35: We should switch these to be the open source tools our community is providing. So you will need links either that are:
a) the latest verified build from the tagged repo, e.g. for make this would be: https://github.com/ZOSOpenTools/makeport/releases/tag/makeport_306
b) a general link to where the tagged builds are and they can choose the one they want: e.g. for make this would be: https://github.com/ZOSOpenTools/makeport/tags
c) the latest 'boot' tagged build which is what they would get if they used zopen-setup, e.g. for make this would be: https://github.com/ZOSOpenTools/makeport/releases/tag/boot

My recommendation would be to get the 'boot' tagged build or the 'latest verified' tagged build (which should be slightly higher quality, or the same as the boot build)

README.md
```diff
@@ -33,20 +32,24 @@ First, you need to have some tools installed on your system:
...
+ - [curl](https://github.com/curl/curl)[^curl-license] - Download unsupported binary from Rocket
+ - [gunzip](https://www.gnu.org/software/gzip/)[^gpl]- Download unsupported binary from Rocket
 
+[^gpl]: [GPL Version 3 or later](https://www.gnu.org/licenses/gpl-3.0.html)
+[^lgpl]: [GNU Lesser Public License 2.1](https://git.kernel.org/pub/scm/git/git.git/tree/LGPL-2.1)
+[^ibm]: web deliverable
+[^curl-license]: [Curl license](https://github.com/curl/curl/blob/master/COPYING)
 ### Recommended software:
- - bash: Download unsupported binary from Rocket
+ - [bash](https://www.gnu.org/software/bash/)[^gpl] - Download unsupported binary from Rocket
```
MikeFultonDev, line 46: should be the z/os community bash as above

MikeFultonDev (changes requested):
just need to switch the links over to the z/os open tools releases and should be good

README.md
ijmitch: right, the Rocket-sourced binaries seem less easy to get hold of than they used to be.

I started this PR after thinking it should be easier to recognize the licenses of the underlying code which the ports use as their starting point. That's why I was linking back to the project pages and making footnotes for which license each one adheres to.

So the mods to the list entries were more about the original projects than the porting repos. I notice that the contributing guidelines (eg [these](https://github.com/ZOSOpenTools/gitport/blob/main/CONTRIBUTING.md)) say:

> Make sure to add the original project's LICENSE file within the patches directory

but I'm not seeing that in practice (eg nothing in [here](https://github.com/ZOSOpenTools/makeport/tree/main/patches)). Or am I misinterpreting that?

In here, how about a table rather than a list? Could have columns for *Project name* (as a link to its main page), *license* and *Port Repo* (again as a link).

README.md
```diff
@@ -58,12 +63,15 @@ The tools have dependencies on other tools, and there are also typically 2 ways
  - one that is not pre-reconfigured and therefore does require autoconf/automake and associated tools
 
 To build from scratch, start with the tarballs of the following tools:
- - m4: requires m4, curl in boot and xlclang installed on the system.
- - perl: additionally requires make, git in boot and m4 in prod.
- - make: additionally requires perl in prod for running test cases.
- - libz: additional requires make in prod
```
ijmitch, line 64: @MikeFultonDev ... I presume this should be _zlib_ ?

README.md
MikeFultonDev: Ah. Ok. Each of our repositories should have an Apache license for the repo and if we have patches then the patches directory should have the license associated with the associated repo. So for example our gitport patches directory should have a license file that is the same as the associated git repo. If we have missed one or it’s the wrong license in a patches directory then that is a bug

README.md
MikeFultonDev: Yes.

ijmitch: corrected

README.md
ijmitch: I create #103 to document which of the repos seem have patches directories without any license file.

MikeFultonDev (approved):
Thanks Ian - looks great

PR #106, opened 2023-01-10
Add the actual dependent projects in the table "Projects with the most dependencies"

IgorTodorovskiIBM:


tools/getbinaries.py
```diff
@@ -93,7 +93,7 @@
 			dependencies += match.split();
 		dependencies = list(set(dependencies))
 		for x in dependencies:	
-			dependentOn[x + "port"] += 1
+			dependentOn[x + "port"] += [name]
```
MikeFultonDev, line 96: So you want to list all dependencies or the first N?

IgorTodorovskiIBM: The table was still readable with all, so I went ahead with all dependencies. :)

MikeFultonDev (approved):

PR #127, opened 2023-01-19
Updates to zopen download, plus zopen update|upgrade

IgorTodorovskiIBM:
Implements @dougburns suggestions:
```
"zopen download" with no parameters implicitly does a --list
"zopen download --update" updates existing packages
"zopen download --all" gets everything, with update of anything already there
"zopen download <packagelist>" downloads the specified packages, updating any of those that already exist
```

In addition, it aliases `zopen update` and `zopen upgrade` to `zopen download --update` 
Also adds `zopen init` description to the zopen help

### DIFF
--- a/bin/lib/zopen-download
+++ b/bin/lib/zopen-download
@@ -8,12 +8,12 @@ export utildir="$( cd "$(dirname "$0")" >/dev/null 2>&1 && pwd -P )"
 printSyntax() 
 {
   args=$*
-  echo "zopen-download is a download utility for z/OS Open Tools. The default action is to download/update all packages." >&2
+  echo "zopen-download is a download utility for z/OS Open Tools. The default action is to download/upgrade all packages." >&2
   echo "If you have a Github OAUTH token, export the environment variable ZOPEN_GIT_OAUTH_TOKEN" >&2
   echo "Syntax: zopen-download [<option>]*" >&2
   echo "  where <option> may be one or more of:" >&2
   echo "  --list: list all available z/OS Open Tools. If no parameters are specified, --list is the default."  >&2
-  echo "  -u|--update|--upgrade: updates installed z/OS Open Tools packages."  >&2
+  echo "  -u|--upgrade: upgrades installed z/OS Open Tools packages."  >&2
   echo "  --reinstall: reinstall already installed z/OS Open Tools packages."  >&2
   echo "  --all: downloads all z/OS Open Tools packages."  >&2
   echo "  -v: run in verbose mode." >&2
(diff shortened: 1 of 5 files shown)


AnthonyGiorgio (approved):

bin/lib/zopen-download
```diff
@@ -12,13 +12,14 @@ printSyntax()
   echo "If you have a Github OAUTH token, export the environment variable ZOPEN_GIT_OAUTH_TOKEN" >&2
   echo "Syntax: zopen-download [<option>]*" >&2
   echo "  where <option> may be one or more of:" >&2
-  echo "  --list: list all available z/OS Open Tools"  >&2
-  echo "  --update: update z/OS Open Tools packages. On by default."  >&2
-  echo "  --force: force downloads all z/OS Open Tools packages."  >&2
+  echo "  --list: list all available z/OS Open Tools. If no parameters are specified, --list is the default."  >&2
+  echo "  -u|--update|--upgrade: updates installed z/OS Open Tools packages."  >&2
```
MikeFultonDev, line 16: I know with Brew that upgrade and update mean different things. Can we just go with the one that brew used (I think it is update but not positive)

MikeFultonDev (commented):
lol good except for one comment

bin/lib/zopen-download
IgorTodorovskiIBM: The brew upgrade command is used to upgrade installed packages to the latest version. Same goes for `apt-get upgrade`. `brew update`, `apt-get update` gets the latest package lists and doesn't actually update anything. 

We could introduce a similar option, but currently we always fetch the package info directly from github.

However, as a future improvement, we could cache the package info from github and use `zopen update` to update the cache.

MikeFultonDev: So are you thinking just have upgrade for now and we could consider update in the future?

IgorTodorovskiIBM: Yeah, I'll remove update to avoid confusion

IgorTodorovskiIBM:
> Is a similar change needed for setup.sh ?

I just made a similar change, had to create a .installing file to avoid an infinite loop when sourcing the .env in setup.sh. If you know of a cleaner solution, let me know!

MikeFultonDev (approved):

dougburns:
Functionally looks good.

The first line of the command help says "The default action is to download/upgrade all packages." which is not true any longer. The later text says that --list is the default, so you can probably just remove that sentence.

The docs section for zopen download has a similar statement about the default behaviour and might also need some changes to reflect the new options.

IgorTodorovskiIBM:
> Functionally looks good.
> 
> The first line of the command help says "The default action is to download/upgrade all packages." which is not true any longer. The later text says that --list is the default, so you can probably just remove that sentence.
> 
> The docs section for zopen download has a similar statement about the default behaviour and might also need some changes to reflect the new options.

Good eye! I've updated them

dougburns (approved):
lgtm

PR #128, opened 2023-01-20
provide cacert support for metaport, git, curl

MikeFultonDev:
add a 'append_to_setup' optional function capability and also move 'touch' of .installed to always happen 
(even when not hardcoded paths) so that setup.sh doesn't run more than once (important if append_to_setup is provided)


IgorTodorovskiIBM (approved):
LGTM

PR #129, opened 2023-01-21
Support OAUTH and provide better error messages for zopen-setup

MikeFultonDev:
Now zopen-setup will give a nice diagnostic for either a 401 (unauthorized) or 403 (forbidden). 
A 401 can occur if an invalid or expired OAUTH credential is provided.
A 403 can occur if the number of attempts to download from github is exceeded and there is no OAUTH id set up.


IgorTodorovskiIBM (approved):
LGTM

PR #131, opened 2023-01-22
change order of tools to be set up from .bootenv and fixed UX issues

MikeFultonDev:
Added graceful failure when package not found (404) so that we can use zopen-setup while we add more boot releases

Cleaned up code that was cleaning up the directories before creating symbolic links

Changed the order of the packages so 'meta' is first and added in more packages

### DIFF
--- a/tools/src/zopen_boot_uri.h
+++ b/tools/src/zopen_boot_uri.h
@@ -6,5 +6,11 @@
   #define ZOPEN_BOOT_URI_PREFIX "ZOSOpenTools"
   #define ZOPEN_BOOT_URI_SUFFIX "releases/download/boot"
   #define ZOPEN_BOOT_URI_TYPE "zos.pax.Z"
-  #define ZOPEN_BOOT_PKG { "curl", "gzip", "tar", "make", "git", "meta", NULL }
+
+  /*
+   * Make sure 'meta' is first - we should make all packages independent, but they are not quite yet
+   * We may want to trim down the list of tools a bit. In particular vim and ncurses are not _required_
+   * but they sure are nice.
+   */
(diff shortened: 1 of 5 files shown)


tools/src/zopen_boot_uri.h
```diff
@@ -6,5 +6,11 @@
...
   #define ZOPEN_BOOT_URI_SUFFIX "releases/download/boot"
   #define ZOPEN_BOOT_URI_TYPE "zos.pax.Z"
-  #define ZOPEN_BOOT_PKG { "curl", "gzip", "tar", "make", "git", "meta", NULL }
+
+  /*
+   * Make sure 'meta' is first - we should make all packages independent, but they are not quite yet
+   * We may want to trim down the list of tools a bit. In particular vim and ncurses are not _required_
+   * but they sure are nice.
+   */
+  #define ZOPEN_BOOT_PKG { "meta", "curl", "make", "git", "less", "perl", "jq", "bash", "diffutils", "findutils", "coreutils", "tar", "gzip", "xz", "bzip2", "vim", "ncurses", NULL }
```
IgorTodorovskiIBM, line 15: Should we add coreutils so that we can eliminate asciiecho: https://github.com/ZOSOpenTools/meta/blob/main/bin/lib/zopen-build#L20?

IgorTodorovskiIBM (approved):
LGTM

tools/src/zopen_boot_uri.h
MikeFultonDev: Coreutils is there

IgorTodorovskiIBM: Ah right!

PR #132, opened 2023-01-22
Add install-or-upgrade option to zopen download to eliminate having to re-download packages every build

IgorTodorovskiIBM:
* Also appends flags to make install command, otherwise make install that has compile steps may fail (happened in Git)


MikeFultonDev (approved):
Lgtm

PR #133, opened 2023-01-23
Add new 'common solution' for S_TYPExxx macros

MikeFultonDev:
- fix broken link to 'Updating docs', 
- add back 'Common Solutions'
- add the new common solution to working around broken S_TYPExxxx macros

### DIFF
--- a/docs/Guides/CommonSolutions.md
+++ b/docs/Guides/CommonSolutions.md
@@ -144,3 +144,49 @@ cp /usr/lpp/cbclib/xlclang/etc/xlclang.cfg /u/fred/xlclang.cfg.zos204.noceertfm
 export CLC_CONFIG=/u/fred/xlclang.cfg.zos204.noceertfm
 ```
 
+## S_TYPEISSHM macro gives an error when compiled
+
+The error might be something like: `called object type 'int' is not a function or function pointer`
+
+This is due to a bug in the LE header file `sys/modes.h` which defines the macros as follows:
+
+```
+       #ifdef __SUSV3_POSIX                                  /*#5@D1A*/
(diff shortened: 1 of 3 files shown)


docs/Guides/CommonSolutions.md
```diff
@@ -144,3 +144,49 @@ cp /usr/lpp/cbclib/xlclang/etc/xlclang.cfg /u/fred/xlclang.cfg.zos204.noceertfm
 export CLC_CONFIG=/u/fred/xlclang.cfg.zos204.noceertfm
 ```
 
+## S_TYPEISSHM macro gives an error when compiled
+
+The error might be something like: `called object type 'int' is not a function or function pointer`
+
+This is due to a bug in the LE header file `sys/modes.h` which defines the macros as follows:
```
IgorTodorovskiIBM, line 151: Can we fix this in zoslib by overriding sys/modes.h?

docs/_sidebar.md
```diff
@@ -7,6 +7,7 @@
   - [Available ports](/Latest.md)
     - [Current Status](/Progress.md)
   - [FAQ](/Guides/FAQ.md)
+  - [Common Solutions](/Guides/CommonSolutions.md)
```
IgorTodorovskiIBM, line 10: Oops, that was probably my fault

docs/Guides/CommonSolutions.md
MikeFultonDev: If we could fix this in zoslib that would be great.

MikeFultonDev:
@IgorTodorovskiIBM ok to merge?

IgorTodorovskiIBM (approved):
LGTM

PR #135, opened 2023-01-24
Move to open source tar in zopen build code

MikeFultonDev:
Leverage two things:
- files are tagged on extract, so no need to manually tag files
- the '-a' option auto-detects what compression was used so we don't have to run a separate tool for decompression

Tested by building php and git.

php was MUCH faster to get the tarball set up than in the past.

### DIFF
--- a/bin/lib/zopen-build
+++ b/bin/lib/zopen-build
@@ -626,16 +626,11 @@ setEnv()
 tagBinaryFiles()
 {
   absdir="$1"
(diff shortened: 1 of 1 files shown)


bin/lib/zopen-build
```diff
@@ -688,35 +682,23 @@ extractTarBall()
...
+  if [ ${v} -lt 1 ] || [ ${r} -lt 34 ]; then
+    printError "Need to be running at least tar 1.34"
   fi
-  rm -f "${tarball}"
 
-  # tar will incorrectly tag files as 1047, so just clear the tags
-  chtag -R -r "${dir}"
+  printInfo "Extract tarball ${tarballz} into ${dir}"
+
+  tar -axf "${tarballz}"
```
IgorTodorovskiIBM, line 697: cool! I didn't know about -a

IgorTodorovskiIBM (approved):
Looks good!

bin/lib/zopen-build
MikeFultonDev: Me either. Learned about it when perusing the new man pages.

PR #136, opened 2023-01-24
Use /bin/echo to intrepret backslash escapes. Modern echo requires -e option

IgorTodorovskiIBM:
If we pass -e to /bin/echo it gets displayed. Rather than adding conditions to handle both coreutils echo and /bin/echo, just use /bin/echo.


MikeFultonDev (approved):
Lgtm

PR #139, opened 2023-01-25
Increase buffer size used for GitHub token.

redstar:
According to the [GitHub blog](https://github.blog/changelog/2021-03-31-authentication-token-format-updates-are-generally-available/) a GitHub token can be aup to 255 characters in length. The current temporary buffer has size 80, which is too small.
This PR changes the size to 280 (255 for the token, 20 for Authorization:Bearer, and a couple of spare bytes).


MikeFultonDev (commented):
can you either add a comment on why the magic 280 or alternately do the math in the array, e.g. 
char buf[255+...] with a comment (and ideally a pointer to the github link).

redstar:
Done.

MikeFultonDev (approved):
thanks!

PR #140, opened 2023-01-25
Migrate source to OpenXL

redstar:
Compiling the source with OpenXL results in 2 errors:

- In createdirs.c the include <string.h> is missing
- In download.c, the macro _EXT must be defined for the fldata() function.

There are many warnings coming from OpenXL - this only fixes the compile errors.


IgorTodorovskiIBM (approved):
LGTM

MikeFultonDev (approved):
LGTM!

PR #141, opened 2023-01-26
Add monitor_vpc.sh script

IgorTodorovskiIBM:
* This script is used to help automatically restart or stop the z/OS VPCs

We can add this to a cron job that checks if the systems are down and it can automatically restart them.


MikeFultonDev (commented):
Where does this script run?

IgorTodorovskiIBM:
Linux, it will run in the Jenkins server

MikeFultonDev (approved):
Lgtm

PR #144, opened 2023-01-29
Move setup.sh to end of .env so that you can use envars in setup.sh

IgorTodorovskiIBM:


MikeFultonDev (approved):
I think it’s good - can you check that the append-to-setup works ok still?

IgorTodorovskiIBM:
> I think it’s good - can you check that the append-to-setup works ok still?

Just noticed "Setup completed" is emitted before that, so I'll fix that up.

IgorTodorovskiIBM:
Ok looks good now, tested with gitport

PR #147, opened 2023-01-30
Delete existing install directory before extracting + Address binary files in .gitattributes + Use default UTF-8 encoding

IgorTodorovskiIBM:
Two parts.
### Zopen download issue:
This addresses: https://github.com/ZOSOpenTools/bashport/issues/40
It also addresses the issue when a .installed file remains after a redownload

### Zopen build .gitattributes issue:
* Issue: https://github.com/ZOSOpenTools/zstdport/issues/6
* Also enforces that Git 2.39 is used.
* Also changes the .gitattributes so that the default encoding of UTF-8 is used (files will be tagged appropriately still)
* Also adds binary tagged files to .gitattributes

### DIFF
--- a/bin/lib/zopen-build
+++ b/bin/lib/zopen-build
@@ -649,8 +649,8 @@ gitClone()
   v=${gitraw%%.*}
   vr=${gitraw%.*}
   r=${vr##*.}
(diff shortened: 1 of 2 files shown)


MikeFultonDev (approved):
LGTM

PR #149, opened 2023-01-31
Add git as an implicit dep since we use it to initialize the tar source dir as a git repo

IgorTodorovskiIBM:
* To ensure that we don't user's git, but the z/OS Open Tools git
* Also check the version for git on both tar and source builds


MikeFultonDev (approved):
LGTM - so we always need git because it is used even for tarball builds for local patching i'm assuming.

PR #150, opened 2023-02-01
Add support for $HOME/zopen as target location and running without _BPX_AUTOCVT set

MikeFultonDev:
This PR fixes two bugs:
- if someone does not have _BPX_AUTOCVT=ON set, then errors occur - fix is to explicitly set the environment variable when running a shell that cares
- If someone specifies a target location of $HOME/zopen it will now 'work' (writing to that location and not creating a symlink)

### DIFF
--- a/tools/src/ceeuopt.s
+++ b/tools/src/ceeuopt.s
@@ -1,0 +1,77 @@
+*/****************************************************************/     00010000
+*/* LICENSED MATERIALS - PROPERTY OF IBM                         */     00020000
+*/*                                                              */     00030000
+*/* 5650-ZOS                                                     */     00040000
+*/*                                                              */     00050000
+*/*     COPYRIGHT IBM CORP. 1991, 2012                           */     00060000
+*/*                                                              */     00070000
+*/* US GOVERNMENT USERS RESTRICTED RIGHTS - USE,                 */     00080000
+*/* DUPLICATION OR DISCLOSURE RESTRICTED BY GSA ADP              */     00090000
+*/* SCHEDULE CONTRACT WITH IBM CORP.                             */     00100000
+*/*                                                              */     00110000
+*/* STATUS = HLE7790                                             */     00120000
+*/****************************************************************/     00130000
+CEEUOPT  CSECT                                                          00180000
+CEEUOPT  AMODE ANY                                                      00190000
+CEEUOPT  RMODE ANY                                                      00200000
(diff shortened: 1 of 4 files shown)


tools/src/ceeuopt.s
```diff
@@ -0,0 +1,77 @@
...
+               STACK=(128K,128K,ANYWHERE,KEEP,512K,128K),              X00680000
+               STORAGE=(NONE,NONE,NONE,0K),                            X00690000
+               TERMTHDACT=(TRACE,,96),                                 X00700000
+               THREADHEAP=(4K,4K,ANYWHERE,KEEP),                       X00710000
+               THREADSTACK=(OFF,4K,4K,ANYWHERE,KEEP,128K,128K),        X00720000
+               TRACE=(OFF,4K,DUMP,LE=0),                               X00730000
+               TRAP=(ON,SPIE),                                         X00740000
+               UPSI=(00000000),                                        X00750000
+               VCTRSAVE=(OFF),                                         X00760000
+               XPLINK=(OFF),                                           X00780000
```
IgorTodorovskiIBM, line 75: XPLINK is off?

Do we need all of these settings or just the FILETAG one?

tools/src/zopensetupmain.c
```diff
@@ -43,11 +43,22 @@ static void syntax(const char* pgm) {
...
                   " -v : print out verbose messages\n"
-                  " -q : only print out errors\n",
+                  " -q : only print out errors\n"
+                  "Note:\n"
+                  " Special consideration is made if <root> is ${HOME}/zopen in which case no link is created\n",
                   pgm, ZOPEN_TOOLS_URL, pgm);
   return;
 }
 
+#define _CVTSTATE_OFF     0
```
IgorTodorovskiIBM, line 53: Can you include _Nascii.h instead or it only works with -qASCII?

tools/src/ceeuopt.s
MikeFultonDev: These settings should be all the defaults - so it's a little verbose but is kind of useful so you see all the options.
Note that this code is built EBCDIC, 31-bit since it needs to work with the web toolkit code that is like this.

tools/src/zopensetupmain.c
MikeFultonDev: it only works _without_ -qASCII... I used the macros just because that's what was in the book. Wasn't sure if _Nascii.h was an external header we could use?

PR #151, opened 2023-02-01
Step 1 of deps parsing of optional operator and version, and cleanup of envars

IgorTodorovskiIBM:
This is step of parsing the dependencies which can have a comparison and version. I am not currently using it in any way yet, that logic will come in a later commit.

I will likely move it to common.inc in step 2 because zopen download will need it for runtime dependencies.

I have also added cleanup_envars, which will cleanup dupes in our envars.

If you have any comments on the code so far, please let me know.

### DIFF
--- a/bin/lib/zopen-build
+++ b/bin/lib/zopen-build
@@ -394,6 +394,35 @@ checkEnv()
   fi
 }
 
+parseDeps()
+{
+  dep="$1"
+  version=$(echo $dep | awk -F '[>=<]+' '{print $2}')
+  if [ -z "$version" ]; then
+    operator=""
+    dep=$(echo $dep | awk -F '[>=<]+' '{print $1}')
+  else
+    operator=$(echo $dep | awk -F '[0-9.]+' '{print $1}' | awk -F '^[a-zA-Z]+' '{print $2}')
+    dep=$(echo $dep | awk -F '[>=<]+' '{print $1}')
+    case $operator in
+      ">=") ;;
+      "=") ;;
+      *) printError "$operator is not supported."
+    esac
+    major=$(echo $version | awk -F. '{print $1}')
+    minor=$(echo $version | awk -F. '{print $2}')
+    if [ -z "$minor" ]; then
+      patch=0
+    fi
+    patch=$(echo $version | awk -F. '{print $3}')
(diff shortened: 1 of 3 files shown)


bin/lib/zopen-build
```diff
@@ -1093,22 +1129,34 @@ if ! [ -f ./.env ]; then
   echo "Need to source from the .env directory" >&2
   return 0
 fi
-export _BPXK_AUTOCVT=ON
-export _CEE_RUNOPTS="\$_CEE_RUNOPTS FILETAG(AUTOCVT,AUTOTAG) POSIX(ON)"
+
+cleanup_envars()
```
MikeFultonDev, line 1133: Does cleanup_envars delete duplicate entries? Is there a better name?

MikeFultonDev (commented):
Just a comment on the dubxtion

MikeFultonDev (approved):

IgorTodorovskiIBM:
@MikeFultonDev, some more details on this PR:

* You can now check for version requirements to the deps:
```
export ZOPEN_TARBALL_DEPS="make>=4.4.0.90"
```
This now checks in the dependency directory for a .version file. The .version file is product via a `zopen_get_version` function which needs to be defined in the buildenv. Right now it will print a warning if it's not defined.

It also introduces a new envar:
```
export ZOPEN_RUNTIME_DEPS="bash>=5.2.0"
```
This info is then copied into a $ZOPEN_INSTALL_DIR/.deps file.

**Step 2** will be to modify Jenkins so that the .version and .deps files are processed and metadata is placed into the github releases. Then zopen download will be modified to extract the version and runtime deps metadata, which will be used for runtime dependency processing.

I tested this PR with a few ports to ensure no regressions.

We'll need to go through and add ZOPEN_RUNTIME_DEPS and zopen_get_version to all of the ports.

MikeFultonDev (approved):
Nice. Should the tool to generate a new port be updated to include templates for this new function

IgorTodorovskiIBM:
> Nice. Should the tool to generate a new port be updated to include templates for this new function

Oh yes, I forgot to do that!

IgorTodorovskiIBM:
Added the templates to zopen generate

MikeFultonDev (approved):

PR #153, opened 2023-02-01
Add install option, also fix zopen download so that it sets up ports

IgorTodorovskiIBM:
The install option is added to make zopen more similar to brew and apt-get.

In addition, as @ccw-1 requested, install should setup the port. Since not all ports have a setup.sh, we'll source the .env for now.

### DIFF
--- a/bin/lib/zopen-build
+++ b/bin/lib/zopen-build
@@ -188,7 +188,7 @@ printSyntax()
   echo "  -h: print this information"
   echo "  -v: run in verbose mode"
   echo "  -vv: run in very verbose mode (sets environment variables V=1 and VERBOSE=1)"
-  echo "  -u|--upgradedeps: upgrade all dependencies by running zopen download"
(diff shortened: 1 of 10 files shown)


MikeFultonDev (commented):
Is install a synonym for download now?

IgorTodorovskiIBM:
Yeah, I'm thinking we can deprecate `zopen download` or use it just to download and not set up.

MikeFultonDev (approved):
I like that. Any reason not to change that asap and switch download -> install? I am not sure there is much need for a download without install?

IgorTodorovskiIBM:
Would it be better to issue a deprecation notice when they use the zopen download command, instead removing it completely? We'll need to make sure we change all references from `zopen download` to `zopen install` and post it on the slack channels/discussions.

MikeFultonDev:
Twerp - agreed

MikeFultonDev (approved):
LGTM

PR #154, opened 2023-02-02
Use /bin/iconv which converts EBCDIC

IgorTodorovskiIBM:
* This is in case libiconv is included as a dependency, which does not support ebcdic currently.


MikeFultonDev (approved):
LGTM
I think the open source iconv supports EBCDIC when built extended but not dbcs EBCDIC. But there might be a bug. This is a good fix

PR #155, opened 2023-02-03
Improve zopen update cacert

MikeFultonDev:
- add complete syntax
- clean up default directory to remove ../ entries
- add support for an alternate location to write the cacert (to be used by git and curl)
- remove setting of environment variables for GIT and CURL (now done in the GIT and CURL .env code)


IgorTodorovskiIBM (approved):
LGTM!

PR #157, opened 2023-02-04
update prefix on install

MikeFultonDev:
Add an additional update to the hardcoded values in the installed code, as follows:

- keep the update so that ZOPEN_INSTALL_ROOT replaced that hard-coded product installation directory
- add a new update so that ZOPEN_INSTALL_PREFIX replaces hard-code prefix locations (e.g. to another pre-req tool)

On install, have setup now:
- update ZOPEN_INSTALL_ROOT to the location the code is now installed into
- update ZOPEN_INSTALL_PREFIX to the parent location of where the code is installed into

This was tested with pipelineport which previously had a '/jenkins/prod' prefix left around causing problems.
This should clean up other tools as well and mean we can eliminate some environment variables I hope (for things like autoconf)


IgorTodorovskiIBM (approved):
Nice! LGTM

PR #163, opened 2023-02-06
Add gonumber + add error checking in update-cacert

IgorTodorovskiIBM:
So that we can get prettier exception messages

Also adds error checking in update-cacert


MikeFultonDev (approved):
LGTM

PR #169, opened 2023-02-09
Various zopen-build Improvements

IgorTodorovskiIBM:
* Doing dir=$(gitClone), doesnt allow gitClone to have any side effects on the variables, so setting `forceRebuild=true` had no effect.
  * Since we set dir within gitClone and downloadTarBall, we don't have to set it to the return value, it's already set.
* Fixes various issues with deleteDuplicateEntries
* Also sorts + uniqs the builddeps and runtime dependencies

### DIFF
--- a/bin/lib/zopen-build
+++ b/bin/lib/zopen-build
@@ -1137,18 +1137,18 @@ deleteDuplicateEntries()
 {
   value=\$1
   delim=\$2
-  echo "\$value\$delim" | awk -v RS="\$delim" '!(\$0 in a) {a[\$0]; printf("%s%s", col, \$0); col=RS; }'
+  echo "\$value\$delim" | awk -v RS="\$delim" '!(\$0 in a) {a[\$0]; printf("%s%s", col, \$0); col=RS; }' | sed "s/\${delim}$//"
 }
 
 _CEE_RUNOPTS="\$_CEE_RUNOPTS FILETAG(AUTOCVT,AUTOTAG) POSIX(ON)"
 # Remove spaces in between brackets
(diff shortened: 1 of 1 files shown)


bin/lib/zopen-build
```diff
@@ -1140,29 +1137,31 @@ deleteDuplicateEntries()
 {
   value=\$1
   delim=\$2
-  echo "\$value" | awk -v RS="\$delim" '!(\$0 in a) {a[\$0]; printf("%s%s", col, \$0); col=RS; }'
+  echo "\$value\$delim" | awk -v RS="\$delim" '!(\$0 in a) {a[\$0]; printf("%s%s", col, \$0); col=RS; }' | sed "s/\${delim}$//"
 }
 
 _CEE_RUNOPTS="\$_CEE_RUNOPTS FILETAG(AUTOCVT,AUTOTAG) POSIX(ON)"
+# Remove spaces in between brackets
+_CEE_RUNOPTS="\$(echo "\$_CEE_RUNOPTS" | awk '{gsub(/\([ \t]*/, "("); gsub(/[ \t]*\)/, ")"); gsub(/[ \t]*,[ \t]*/, ","); print}')"
```
MikeFultonDev, line 1145: this is ok for now, but some runtime options have strings that you wouldn't want to remove spaces from

MikeFultonDev (approved):
LGTM - note my one comment

bin/lib/zopen-build
IgorTodorovskiIBM: I see, USRHDLR(lmname,"") seems to be one example. As long as there is not comma or () within the quotes, then it should be ok.

Issue #162, opened 2023-02-06
make `zopen install`'ed packages relocatable or preserve `download`?


covener:
Relative to this announcement about `zopen download` deprecation:

> Hi, please be aware that zopen download has been deprecated and will be replaced with zopen install. The key difference between these two options is that zopen install not only downloads and extracts the tool, but also sets it up for use, whereas zopen download only performed the first two actions.

I think one thing to consider for preserving `zopen download` is that for many packages it will bake in paths during the installation (sourcing of env.sh) This means the package directory cannot be moved.   The alternative would be grabbing the packages from their individual releases, but as time goes on you might be foregoing more and more package management features.

Beyond preserving download, another option is to make it so that the sourced env scripts can be re-run without looking for their original jenkins prefix.

> Good point! would you be able to open an issue? We might be able to handle it in a different way, as in we could make setup.sh smarter, it could probably store the initial install path somewhere and then on rerun of setup.sh update hardcoded references of that install path to the new one (if you they don't match).

MikeFultonDev:
I think much of this has been fixed. @covener please re-open with details if you see remaining issues.

PR #170, opened 2023-02-09
.env sourcing of dependencies + download-only option

IgorTodorovskiIBM:
* When dependencies are installed, adjust the .env so that it also sources the dependency's .env files.
* Also add a download-only option (like in apt-get install --download-only) to partly address https://github.com/ZOSOpenTools/meta/issues/162

### DIFF
--- a/bin/lib/zopen-install
+++ b/bin/lib/zopen-install
@@ -41,7 +41,8 @@ printSyntax()
   echo "  -u|--upgrade: upgrades installed z/OS Open Tools packages."  >&2
   echo "  --install-or-upgrade: installs the package if not installed, or upgrades the package if installed."  >&2
   echo "  --reinstall: reinstall already installed z/OS Open Tools packages."  >&2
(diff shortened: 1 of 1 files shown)


bin/lib/zopen-install
```diff
@@ -114,14 +115,21 @@ installDependencies()
   if [ "$dependencies" != "No dependencies" ]; then
     printHeader "Installing dependencies for $name: $dependencies..."
     
+    # Remove an old one if it exists
+    rm -f "${name}/.depsenv"
```
MikeFultonDev, line 119: since you have an rm -f here, do you want to protect this with a check it isn't empty and fail?

MikeFultonDev (approved):
LGTM - i had one comment

bin/lib/zopen-install
IgorTodorovskiIBM: I decided not to fail if it's not empty because we could be reinstalling the port. So I forcefully remove it (which won't complain if it doesn't exist).

PR #181, opened 2023-02-13
use pushd/popd when in bash

IgorTodorovskiIBM:
To address https://github.com/ZOSOpenTools/gitport/issues/54

(FYI, in a future commit, I will be moving the deps code to zopen-build)

### DIFF
--- a/bin/lib/zopen-install
+++ b/bin/lib/zopen-install
@@ -120,7 +120,13 @@ installDependencies()
     echo "$dependencies" | xargs | tr ' ' '\n' | sort | while read dep; do
       cat <<ZZ >> "${name}/.depsenv"
 if [ -f "../${dep}/.env" ]; then
-  cd "../${dep}" && . ./.env; cd -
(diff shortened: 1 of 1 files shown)


bin/lib/zopen-install
```diff
@@ -120,7 +120,13 @@ installDependencies()
     echo "$dependencies" | xargs | tr ' ' '\n' | sort | while read dep; do
       cat <<ZZ >> "${name}/.depsenv"
 if [ -f "../${dep}/.env" ]; then
-  cd "../${dep}" && . ./.env; cd -
+  if [[ \$(type echo) == 'echo is a shell builtin' ]]; then
```
MikeFultonDev, line 123: Looks good.

MikeFultonDev (approved):

bin/lib/zopen-install
ccw-1: may be pushd > /dev/null and popd >/dev/null to prevent the print out.

PR #182, opened 2023-02-13
Remove build timestamp in resulting directory for new and existing pax.Z packages

IgorTodorovskiIBM:
* This also modifies existing pax.Z packages that extract to a name.version.timestamp.zos directory and modifes them to name.version.zos.


MikeFultonDev (approved):
Lgtm

PR #201, opened 2023-02-20
work around bug in tar setting time for files

MikeFultonDev:
add the -m option to untar files to work around bug in tar. 
Issue being tracked is: https://github.com/ZOSOpenTools/tarport/issues/8


MikeFultonDev:
@IgorTodorovskiIBM i force merged this since i know you are out

PR #206, opened 2023-02-24
fix: use main as name for primary branch on git init

### LINKED ISSUE #205: Tarball builds get big warning from git init about use of "master" as the primary branch
e.g.
```
Extract tarball findutils-4.9.0.tar.xz into findutils-4.9.0
hint: Using 'master' as the name for the initial branch. This default branch name
hint: is subject to change. To configure the initial branch name to use in all
hint: of your new repositories, which will suppress this warning, call:
hint: 
hint: 	git config --global init.defaultBranch <name>
hint: 
hint: Names commonly chosen instead of 'master' are 'main', 'trunk' and
hint: 'development'. The just-created branch can be renamed via this command:
hint: 
hint: 	git branch -m <name>
```

I think we should be following best practice when creating local git repositories and explicitly request "main".

dougburns:
fixes #205

### DIFF
--- a/bin/lib/zopen-build
+++ b/bin/lib/zopen-build
@@ -742,11 +742,11 @@ extractTarBall()
   # Perhaps this can be done with a function but be aware of blanks, newlines, etc. in file names
   #
   if [ "${ZOPEN_GIT_SETUP}x" = "Nx" ]; then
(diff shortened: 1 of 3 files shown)


IgorTodorovskiIBM (approved):
Nice! LGTM

MikeFultonDev:
nice. lgtm.

PR #211, opened 2023-03-01
Use tar -axf now that its issues are resolved

IgorTodorovskiIBM:
* Remove unnecessary check for utime error


MikeFultonDev (commented):
approved 😁

PR #213, opened 2023-03-01
Various fixes/changes to boottool.sh

IgorTodorovskiIBM:
* Added Authorization: Bearer $GITHUB_OAUTH_TOKEN in some missing places
* Update boot to the latest release so that we don't have to dig it out manually
* Also a convenience script tools/update_boots.sh to update a set of tools automatically.

### DIFF
--- a/tools/boottool.sh
+++ b/tools/boottool.sh
@@ -107,7 +107,7 @@ populateReleaseMaps()
   name=""
   releaseNameIDMap=()
   releaseIDToReleaseNameMap=()
-  response=$(curl -sw "%{http_code}" -k -H "Accept: application/vnd.github+json" $url)
+  response=$(curl -sw "%{http_code}" -k -H "Accept: application/vnd.github+json" -H "Authorization: Bearer $GITHUB_OAUTH_TOKEN" $url)
   http_code=$(tail -n1 <<< "$response")
 
   if [ "$http_code" == "$CURL_REPO_SUCCESS" ]; then
@@ -213,7 +213,7 @@ downloadAssetsOfRelease()
   assetNameArray=()
   assetUrl=$url"/""$1""/""assets"
   #echo "downloadAssetsOfRelease assetsurl = $assetUrl"
-  response=$(curl -sw "%{http_code}" -k -H "Accept: application/vnd.github+json" $assetUrl)
+  response=$(curl -sw "%{http_code}" -k -H "Accept: application/vnd.github+json" -H "Authorization: Bearer $GITHUB_OAUTH_TOKEN" $assetUrl)
   http_code=$(tail -n1 <<< "$response")
 
   if [ "$http_code" == "$CURL_REPO_SUCCESS" ]; then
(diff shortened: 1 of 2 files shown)


tools/boottool.sh
```diff
@@ -46,19 +51,40 @@ processOptions()
   echo "Repo=$REPO"
 }
 
+getLatestReleaseName()
+{
+  latestRepoUrl="https://api.github.com/repos/$OWNERNAME/$REPO/releases/latest"
+  echo $latestRepoUrl
+  
+  response=$(curl -sw "%{http_code}" \
```
MikeFultonDev, line 59: check for curl failure after this?

tools/boottool.sh
```diff
@@ -81,7 +107,7 @@ populateReleaseMaps()
   name=""
   releaseNameIDMap=()
   releaseIDToReleaseNameMap=()
-  response=$(curl -sw "%{http_code}" -k -H "Accept: application/vnd.github+json" $url)
+  response=$(curl -sw "%{http_code}" -k -H "Accept: application/vnd.github+json" -H "Authorization: Bearer $GITHUB_OAUTH_TOKEN" $url)
```
MikeFultonDev, line 110: check rc of curl here

tools/update_boots.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/sh
+
+# Script to automatically update all or a select number of boot tools
+
+if [ -z "$boottools" ]; then
+  echo "Updating all boot tools"
+  boottools="meta,curl,make,git,less,perl,jq,bash,diffutils,findutils,coreutils,tar,gzip,xz,bzip2,vim,ncurses"
```
MikeFultonDev, line 7: with the changes to use the tar -axf service, do we need gzip, xz, bzip2 in boot? Probably good for this rev but maybe we can remove it next time?

MikeFultonDev (changes requested):
added a few comments

tools/update_boots.sh
MikeFultonDev: Ah

tools/boottool.sh
```diff
@@ -46,19 +51,40 @@ processOptions()
...
 {
   echo "Pass repo/release name -- usage: boottool.sh --repo <repo> --release <releasename> --uname [ownername]" 
+  echo "If --release is not passed, it will use the latest release"
   exit 1;
 }
 
 [ -z "${OWNERNAME}" ] && OWNERNAME=$DEFAULT_OWNER && echo "Using default owner name ZOSOpenTools";
 [ -z "${REPO}" ] && (printSyntax)
-[ -z "${RELEASENAME}" ] && printSyntax
+[ -z "${RELEASENAME}" ] && getLatestReleaseName
```
HarithaIBM, line 87: A question, what is the behavior if latest release fetching fails?

IgorTodorovskiIBM: Added another check if RELEASENAME is not updated by getLatestReleaseName

MikeFultonDev (approved):

HarithaIBM (approved):

PR #229, opened 2023-03-14
Move _TAG_REDIR_* envars before functions

IgorTodorovskiIBM:
deleteDuplicateEntries relies on the TAG_REDIR envars, so we need to move them before the function definition

### DIFF
--- a/bin/lib/zopen-build
+++ b/bin/lib/zopen-build
@@ -1168,6 +1168,10 @@ if ! [ -f ./.env ]; then
   return 0
 fi
 
+export _TAG_REDIR_IN=txt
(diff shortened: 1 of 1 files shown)


MikeFultonDev (commented):
can you explain why this matters? I'm not sure I 'get it'

IgorTodorovskiIBM:
> can you explain why this matters? I'm not sure I 'get it'

If you source the .env under /bin/sh without the TAG_REDIR_* macros set, the CEE_RUNOPTS/etc will have a garbled value.

MikeFultonDev (approved):
Thanks

Issue #231, opened 2023-03-16
replaceHardcodedPaths function fails if permissions to file are read only


IgorTodorovskiIBM:
The error at install for perl:
```
replaceHardcodedPaths 20: install 19: /home/itodoro/projects/meta/bin/lib/zopen-build 1648: FSUM7343 cannot open "/home/itodoro/zopen/prod/perl5-blead/lib/5.37.10/os390/Config_heavy.pl.tmp" for output: EDC5111I Permission denied.
```

```
[ITODORO@ZOSCAN2B ~/projects/perlport_new]$ ls -l /home/itodoro/zopen/prod/perl5-blead/lib/5.37.10/os390/Config_heavy.pl
-r--r--r--   1 ITODORO  BASIC      43424 Mar 16 09:50 /home/itodoro/zopen/prod/perl5-blead/lib/5.37.10/os390/Config_heavy.pl
[ITODORO@ZOSCAN2B ~/projects/perlport_new]$ ls -l /home/itodoro/zopen/prod/perl5-blead/lib/5.37.10/os390/Config_heavy.pl.tmp
-r--r--r--   1 ITODORO  BASIC      43424 Mar 16 09:50 /home/itodoro/zopen/prod/perl5-blead/lib/5.37.10/os390/Config_heavy.pl.tmp
```

IgorTodorovskiIBM:
Handled in https://github.com/ZOSOpenTools/meta/pull/234

PR #234, opened 2023-03-16
Handle read-only files when replacing hardcoded install paths

IgorTodorovskiIBM:
* Reduces code duplication by introducing a replaceHardcodedPath function
* Adds code to handle read only files (which generate errors in perl)
Handles https://github.com/ZOSOpenTools/meta/issues/231


MikeFultonDev (approved):
Lgtm!

PR #236, opened 2023-03-17
Add .zoslib_env files

IgorTodorovskiIBM:
Works with zoslib PR: https://github.com/ibmruntimes/zoslib/pull/13.

For git, this allows it to run without having to source the .env:
```
zopen_append_to_zoslib_env() {
cat <<EOF
GIT_MAN_PATH|unset|
GIT_ROOT|unset|
GIT_SHELL|unset|
GIT_TEMPLATE_DIR|set|PROJECT_ROOT/share/git-core/templates
GIT_EXEC_PATH|set|PROJECT_ROOT/libexec/git-core
GIT_SSL_CAINFO|set|PROJECT_ROOT/cacert.pem
EOF
}
```

### DIFF
--- a/bin/lib/zopen-build
+++ b/bin/lib/zopen-build
@@ -91,14 +91,15 @@ User-Provided environment variables:
 printFunctions() 
 {
 echo "User-Provided functions:
-  zopen_init               This function runs after code is downloaded and patched but before the code is built. 
-  zopen_pre_check          This function runs before the 'check' step of the build is run.
-  zopen_append_to_env      This function runs as part of generation of the .env file. The output of the function is appended to .env
-  zopen_append_to_setup    This function runs as part of generation of the setup.sh file. The output of the function is appended to setup.sh
-  zopen_pre_install        This function runs before the 'install' step of the build is run.
-  zopen_post_install       This function runs after the 'install' step of the build is run.
-  zopen_check_results      This function runs after the 'check' step of the build is run and must print out expected and actual failures."
-  zopen_get_version        This function returns the version of the tool in accordance with semantic versioning
+  zopen_init                  This function runs after code is downloaded and patched but before the code is built. 
+  zopen_pre_check             This function runs before the 'check' step of the build is run.
+  zopen_append_to_env         This function runs as part of generation of the .env file. The output of the function is appended to .env
+  zopen_append_to_zoslib_env  This function runs as part of generation of the .zoslib_env file. The output of the function is appended to .zoslib_env
+  zopen_append_to_setup       This function runs as part of generation of the setup.sh file. The output of the function is appended to setup.sh
+  zopen_pre_install           This function runs before the 'install' step of the build is run.
+  zopen_post_install          This function runs after the 'install' step of the build is run.
+  zopen_check_results         This function runs after the 'check' step of the build is run and must print out expected and actual failures."
+  zopen_get_version           This function returns the version of the tool in accordance with semantic versioning
 }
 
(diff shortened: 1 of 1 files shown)


MikeFultonDev (changes requested):
we discussed this and the recommendation is to go with a solution that binds this information into the programs rather than using a side file

IgorTodorovskiIBM:
> we discussed this and the recommendation is to go with a solution that binds this information into the programs rather than using a side file

The latest commits address this. The related zoslib PR is: https://github.com/ibmruntimes/zoslib/pull/13

bin/lib/zopen-build
```diff
@@ -1588,6 +1590,126 @@ LABEL specification=1.0.0"
...
+    cat <<zz >>$output
+    value_str = "$value";
+    size = PATH_MAX;
+    envar_value = (char*)malloc(size);
+    memset(envar_value, 0, size);
+
+    // Substitute PROJECT_ROOT with root_dir
+    pos = strstr(value_str, "PROJECT_ROOT");
+    if (pos != NULL) {
+      int len = pos - value_str;
```
MikeFultonDev, line 1646: maybe something more descriptive than len? This is the length of the prefix before the PROJECT_ROOT correct?

bin/lib/zopen-build
```diff
@@ -1588,6 +1590,126 @@ LABEL specification=1.0.0"
...
+      strncat(envar_value, value_str, len);
+      strcat(envar_value, root_dir);
+      strcat(envar_value, pos + strlen("PROJECT_ROOT"));
```
MikeFultonDev, line 1649: since PROJECT_ROOT is a hard-coded string, why not have a macro with that string and then you can have a PROJECT_ROOT_LEN that is just `sizeof(PROJECT_ROOT)-1 `? That saves a strlen call...

bin/lib/zopen-build
```diff
@@ -1588,6 +1590,126 @@ LABEL specification=1.0.0"
...
+    } else {
+      strcpy(envar_value, value_str);
+    }
+
+#ifdef __ZOPEN_DEBUG
+    fprintf(stderr, "Setting variable %s=%s\n", "$var", envar_value);
+#endif
+    if (!getenv("$var")) {
+      if (setenv("$var", envar_value, 1) != 0) {
+        fprintf(stderr, "Error: Setting environment variable %s=%s\n", "$var", envar_value);
```
MikeFultonDev, line 1659: shouldn't this be a hard failure if the env var can't be set?

bin/lib/zopen-build
```diff
@@ -1588,6 +1590,126 @@ LABEL specification=1.0.0"
...
+      }
+    }
+    free(envar_value);
+zz
+  elif [ "$action" = "prepend" ]; then
+    cat <<zz >>$output
+    value_str = "$value";
```
MikeFultonDev, line 1666: similar comments as for set

bin/lib/zopen-build
```diff
@@ -1588,6 +1590,126 @@ LABEL specification=1.0.0"
...
+zz
+  fi
+done
+
+echo "" >> $output
+echo "return 0;" >> $output
+echo "}" >> $output
+
+if [ "$CC" = "xlclang" ]; then
+  extraOptions="-qexportall"
```
MikeFultonDev, line 1700: Why does xlclang need exportall? Why doesn't clang? Does everything have to be exported, or could you use a `#pragma export` directive ?

MikeFultonDev (changes requested):

MikeFultonDev (approved):
Lgtm

PR #238, opened 2023-03-19
enable alternate shell to be specified

MikeFultonDev:
simple update so that if ZOPEN_SHELL is set, it will use that shell instead of '/bin/sh'.
For me, it lets me run with bash and inherits all the settings like PS1 and set -o emacs and such...


IgorTodorovskiIBM (approved):
LGTM

PR #239, opened 2023-03-19
Have 'git' generate a MANPATH entry

MikeFultonDev:
When zopen-build checks to see if there is a man directory, it won't exist because git builds the man pages using a post-install process.
So - this needs to be overridden with a specific check in the code for 'GIT' as a special package.

This fix (or an alternate) needs to be done before the 'man' support for git PR is done otherwise the generated env scripts won't be right.


IgorTodorovskiIBM (approved):
LGTM, should the shell changes be removed from this PR since they're in another PR?

PR #245, opened 2023-03-22
Add initial support for building with Go

IgorTodorovskiIBM:
* Initial support, testing with [fzfport](https://github.com/ZOSOpenTools/fzfport) 
* Go's envsetup needs to be sourced prior to building.
* TODOs:
 * [ ] Decide on what should the min. Go compiler version be? @ccw-1 any thoughts on this? We currently only have access to V2R4 CI instances.
 * [ ] Any other environment variables that should be checked or enforced?
 * [ ] Other considerations?

### DIFF
--- a/bin/lib/zopen-build
+++ b/bin/lib/zopen-build
@@ -545,6 +545,15 @@ setEnv()
       export ZOPEN_LDFLAGSD="${ZOPEN_CLANG_LDFLAGSD}"
       optflags="-O3"
       ;;
+    GO)
+      export ZOPEN_CCD=""
+      export ZOPEN_CXXD=""
+      export ZOPEN_CPPFLAGSD=""
+      export ZOPEN_CFLAGSD=""
+      export ZOPEN_CXXFLAGSD=""
+      export ZOPEN_LDFLAGSD=""
+      optflags=""
+      ;;
     *)
       printError "Unrecognize compiler '${ZOPEN_COMP}'"
       ;;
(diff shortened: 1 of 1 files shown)


IgorTodorovskiIBM:
Although https://github.com/mikefarah/yq has a Makefile it can also be built with `go`.

We probably also want to add support for `go build`/`go install` and/or use https://github.com/ZOSOpenTools/wharf.

IgorTodorovskiIBM:
Also looking at supporting Go's "build system" - https://github.com/ZOSOpenTools/yqport/blob/main/buildenv as an example.  We may want to consider implementing a new option or variable to toggle the build system. For example:
```
export ZOPEN_BUILD_SYSTEM=Go
```
This would skip configure, use `go build` for "make", use `go test` for test, use `go install` for install. Also potentially run Wharf as a pre-build step.  Further discussions with the Go team will be held to decide on the recommended sequence.

The default build system can be Make as before, but we can also support CMake, Ninja and others which can change the default behaviour and help simplify our buildenv files.

v1gnesh:
Where possible, it may be best to stick with the build methods used by the individual Go projects.

I see `zos` was reserved as a `GOOS` keyword a while ago.
Does that mean the zOS-related parts of Go will be upstreamed eventually, or is that not possible?
Completing `zos` support upstream will mean that we can treat it as just another target (like `s390x`).
Will also simplify upstreaming `zos` compatibility to whichever projects we want to port, as opposed to maintaining zOS-specific patches & build processes here.

IgorTodorovskiIBM:
I will close this because it's now handled via https://github.com/ZOSOpenTools/meta/pull/282 and the comp_go dependency - https://github.com/ZOSOpenTools/comp_go/blob/main/buildenv

PR #260, opened 2023-03-31
-Werror causes issues in the headers since they have warnings

IgorTodorovskiIBM:
Some projects check if Werror is an allowed option and then set it, but our headers generate warnings so this cause the build to fail. So passing -Wno-error.


MikeFultonDev (approved):
unfortunate, but LGTM

PR #275, opened 2023-04-26
Fix broken zopen-setup by adding a cert for objects.githubusercontent.com

IgorTodorovskiIBM:
This is a bit of a hacky/quick fix, we'll most likely phase zopen-setup out with the incoming zopen changes.

This fixes this issue:
```
Download meta into /u/mvm/zopen/boot
*ERROR* rc: 262 ***
toolkit DIAG: Service[1441793], Reason[8], Desc[Certificate validation error]

FATAL ERROR encountered [Connect]
error downloading  https://github.com/ZOSOpenTools/metaport/releases/download/boot/meta-main.20230227_023657.zos.pax.Z to /u/mvm/zopen/boot/meta-main.20230227_023657.zos.pax.Z: -1
error -1 downloading https://github.com/ZOSOpenTools/metaport/releases/download/boot/meta-main.20230227_023657.zos.pax.Z with PEM file /tmp/zopen_tmp_16842817.pem to /u/mvm/zopen/boot/meta-main.20230227_023657.zos.pax.Z
MVM:/u/mvm: >ping zosopentools.github.io                                                                                                                                           
CS V2R5: Pinging host zosopentools.github.io (185.199.110.153) 
Ping #1 response took 0.034 seconds. (33.949 milliseconds)
```

It looks like the redirect to objects.githubusercontent.com uses another ca cert. Adding it fixes the issue


MikeFultonDev (approved):
Hm. ok... if this needs to stick around for more than a couple weeks maybe it would be good to see if we can just concatentate the pem files together rather than 2?

IgorTodorovskiIBM:
> Hm. ok... if this needs to stick around for more than a couple weeks maybe it would be good to see if we can just concatentate the pem files together rather than 2?

I had issues when importing a bundled certificate. I had to split them into individual certs for the gskkyman import to work. If there's a way to import a bundled cert, then yes I agree that would be better.

PR #276, opened 2023-04-27
Create codeql.yml

IgorTodorovskiIBM:
Enable codeql scanning, will try it on meta first

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,76 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+#
+# ******** NOTE ********
+# We have attempted to detect the languages in your repository. Please check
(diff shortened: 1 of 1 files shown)


MikeFultonDev:
can you provide a little info on what it's doing? what types of scans?

MikeFultonDev (commented):
.

IgorTodorovskiIBM:
Hey Mike, it's a security code scanning tool - more info here: https://codeql.github.com/

The one issue right now is that is uses the file extensions to determine the file type. So our .patch files would not be analyzed.

MikeFultonDev:
Hm. That’s too bad. Maybe we should put patch. At the front? Support both styles for search write a script to switch it all, then change the pattern? Back burner item…

IgorTodorovskiIBM:
> Is there a way to say that files with no extension could be a script and to look in the first line (eg buildenv)

Looks like the code needs to be buildable - https://stackoverflow.com/questions/62105456/how-can-i-use-codeql-on-partial-c-source-code.

IgorTodorovskiIBM:
I'll close this for now, we'll need a github action that clones the source and runs the static scanning.

PR #277, opened 2023-04-28
Add ZOPEN_TYPE=BARE for cases where the binaries already exist

IgorTodorovskiIBM:
ZOPEN_TYPE=BARE will avoid downloading and applying patches.

This is one example that leverages https://github.com/ZOSOpenTools/ibm_pythonport/blob/main/buildenv#L13.

Additionally, this adds error checking when sourcing the .env file, so it they return non-zero, zopen-build will print an error.

TODO:
* [ ] Add documentation

### DIFF
--- a/bin/lib/common.inc
+++ b/bin/lib/common.inc
@@ -243,6 +243,7 @@ validateVersion()
 
 initDefaultEnvironment()
 {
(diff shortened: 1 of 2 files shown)


MikeFultonDev (approved):
nice

PR #282, opened 2023-05-01
Refactor and offload most compiler logic to dependencies

IgorTodorovskiIBM:
In some cases, projects may need to depend on multiple compilers, such as Go and C (in the case of fzf). Rather than hardcoding the logic into zopen-build, it may be easier to think of the compilers as yet another dependency.

The logic for finding the compiler and setting the default flags is shifted to the dependency .env file which is generated by the buildenv:

Clang - https://github.com/ZOSOpenTools/comp_clangport/blob/main/buildenv
XL Clang - https://github.com/ZOSOpenTools/comp_xlclangport/blob/main/buildenv
Go - https://github.com/ZOSOpenTools/comp_xlclangport/blob/main/buildenv

For now they just detect the presence of the compiler and if present, they set the default flags. At some point, once Clang and Go are fully upstreamed, we can change the buildenv configuration to build them from source.

If you need to add clang and go as a dependency, you can now do it as follows:
ZOPEN_TARBALL_DEPS="comp_clang comp_go"

### DIFF
--- a/bin/lib/zopen-build
+++ b/bin/lib/zopen-build
@@ -405,7 +405,7 @@ checkEnv()
     ext=${ZOPEN_TARBALL_URL##*.}
     if [ "${ext}x" = "xzx" ]; then
       implicitDeps="${implicitDeps} xz"
-    elif [ "${ext}x" = "gzx" ]; then
+    elif [ "${ext}x" = "gzx" -o  "${ext}x" = "tgzx" ]; then
       implicitDeps="${implicitDeps} gzip"
     elif [ "${ext}x" = "bz2x" ]; then
       implicitDeps="${implicitDeps} bzip2"
@@ -739,6 +739,7 @@ downloadTarBall()
   fi
   tarballz=$(basename "$ZOPEN_TARBALL_URL")
   dir=${tarballz%%.tar.*}
+  dir=${dir%%.tgz}
   if [ -d "${dir}" ]; then
     echo "Using existing tarball directory ${dir}" >&2
   else
@@ -1753,12 +1754,8 @@ if ! resolveCommands; then
 fi
 
 if command -V "${ZOPEN_INIT_CODE}" >/dev/null 2>&1; then
-  if [ -r "${ZOPEN_LOG_DIR}/config.success" ]; then
-    echo "Using previous successful run of ${ZOPEN_INIT_CODE}. Specify -f to re-run" >&2
-  else
-    printVerbose "Running ${ZOPEN_INIT_CODE}"
-    "${ZOPEN_INIT_CODE}" "${PWD}"
-  fi
+  printVerbose "Running ${ZOPEN_INIT_CODE}"
+  "${ZOPEN_INIT_CODE}" "${PWD}"
 fi
 
 if ${startShell}; then
--- a/bin/lib/zopen-build
+++ b/bin/lib/zopen-build
@@ -114,7 +114,7 @@ export utilparentdir="$(cd "$(dirname "$0")/../.." >/dev/null 2>&1 && pwd -P)"
 for tmp in "$TMPDIR" "$TMP" /tmp
 do
   if [ ! -z $tmp ] && [ -d $tmp ]; then
-    TEMP_C_FILE="$tmp/$LOGNAME.c"
+    export ZOPEN_TEMP_C_FILE="$tmp/$LOGNAME.c"
     break
   fi
 done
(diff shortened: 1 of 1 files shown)


bin/lib/zopen-build
```diff
@@ -376,6 +362,25 @@ checkEnv()
...
+    ZOPEN_COMP=$(echo "${ZOPEN_COMP}" | tr '[a-z]' '[A-Z]')
+  fi
+
+  echo "IN ${ZOPEN_COMP}"
+  case "${ZOPEN_COMP}" in 
+    XL)
+      implicitDeps="$implicitDeps ibm_xlclang"  
+    ;;
+    CLANG)
+      implicitDeps="$implicitDeps ibm_clang"
```
perry-ca, line 379: Just call it clang.  We also have the ibm-clang command from OpenXL.  Calling the OEL clang ibm_clang will just confuse things and would make adding ibm-clang support interesting (would it be ibm_ibm-clang :) ). 

Maybe we should change  the --comp option so it specifies the name of the dependency.  This would avoid this mapping step and would allow us to add support  for additional compilers without needing to modify this script.  I'd name all the compiler dependencies comp_XXX.  That way we can easily list all compilers supported in the help and at this point add `comp_${ZOPEN_COMP}` to the implicit dependencies.

IgorTodorovskiIBM: Good suggestions!

bin/lib/zopen-build
```diff
@@ -207,7 +193,7 @@ printSyntax()
   echo "  -vv: run in very verbose mode (sets environment variables V=1 and VERBOSE=1)"
   echo "  -u|--upgradedeps: upgrade all dependencies by running zopen install"
   echo "  --buildtype: release|debug. The default is release"
-  echo "  --comp: xl|clang.  The compiler used for building.  The default is xl."
+  echo "  --comp: <compiler dependency>.  The compiler(s) used for building.  The default is comp_xlclang."
```
perry-ca, line 196:
```suggestion
  echo "  --comp <compiler dependency>:  The compiler(s) used for building.  The default is xlclang."
```
Move the ':' to be clearer about option syntax.  Need to do this for --buildtype too.
Also have zopen-build add the "comp_".

perry-ca:
Do you have an example of the comp_* dependency dir/file?

IgorTodorovskiIBM:
> Do you have an example of the comp_* dependency dir/file?

Hey Sean, you can perform a `zopen install comp_clang`. The logic is in the .env file.

perry-ca (approved):
LGTM

bin/lib/zopen-build
```diff
@@ -113,7 +113,7 @@ export utilparentdir="$(cd "$(dirname "$0")/../.." >/dev/null 2>&1 && pwd -P)"
 for tmp in "$TMPDIR" "$TMP" /tmp
 do
   if [ ! -z $tmp ] && [ -d $tmp ]; then
-    TEMP_C_FILE="$tmp/$LOGNAME.c"
+    export ZOPEN_TEMP_C_FILE="$tmp/$LOGNAME.c"
```
MikeFultonDev, line 116: need a check after this loop that you were able to set ZOPEN_TEMP_C_FILE - otherwise the rm -rf is pretty nasty below...

bin/lib/zopen-build
```diff
@@ -518,123 +500,6 @@ setDepsEnv()
 
 setEnv()
 {
-  # Set up the compiler and option variables
-  # CC - the C compiler
```
MikeFultonDev, line 522: where did all the setenv code go?

MikeFultonDev (commented):
2 comments

bin/lib/zopen-build
IgorTodorovskiIBM: That is in the .env of the comp_* dependencies - https://github.com/ZOSOpenTools/comp_xlclangport/blob/main/buildenv#L32

MikeFultonDev (approved):
Lgtm

v1gnesh:
I noticed that there's metaport, assuming it is to update meta.
However, meta is in `$HOME/zopen/boot/meta`.
How do I update `boot/meta` so that this PR is applied on it?
Don't know if I can do a git pull from within `$HOME/zopen/boot/meta`, as there are some untracked files created by the zopen install process.

IgorTodorovskiIBM:
We remove the `.git` directory from meta in the metaport buildenv: https://github.com/ZOSOpenTools/metaport/blob/main/buildenv#L69

But perhaps we shouldn't. It would make pulling changes easier. Currently, you'll need to clone meta somewhere else, such as in your `zopen/dev` and then source the .env under meta.

perry-ca:
+1 to keeping .git.  I'd review why we are deleting any of those files/dirs.

PR #289, opened 2023-05-05
docs: added hint for environment variable

sabitor:
Without having the _BPXK_AUTOCVT environment variable set, the .bootenv script execution will stop because some of the files used by the .bootenv script are in ASCII and tagged as ASCII.
The addvice to ensure that _BPXK_AUTOCVT has to be set to ON would prevent these problems.


MikeFultonDev (approved):
Thanks - good update.

PR #299, opened 2023-05-16
Allow an option to override default zopen type

IgorTodorovskiIBM:
This is useful for when we want to build both from the official tarball release and git release.


MikeFultonDev (approved):

MikeFultonDev:
note that some may only have one or the other (git or tar) so there could be failures if automated.

PR #307, opened 2023-05-24
Update zopen-generate

v1gnesh:
Update `.gitignore` to not track the soon-to-be-cloned project.
Ex: For a project named `goof`, if `zopen generate` creates `goofport`, this one-line addition will add `goof/` to the `.gitignore` so that the pesky +1 in the end goes away.

snip

### DIFF
--- a/bin/lib/zopen-generate
+++ b/bin/lib/zopen-generate
@@ -172,6 +172,7 @@ touch "${name}port/.gitignore" && chtag -tc 819 "${name}port/.gitignore"
 cat <<EOT > "${name}port/.gitignore"
 log/
 install/
+${name}/
 EOT
 printInfo "${name}port/.gitignore created"
 
--- a/bin/lib/zopen-generate
+++ b/bin/lib/zopen-generate
@@ -172,7 +172,7 @@ touch "${name}port/.gitignore" && chtag -tc 819 "${name}port/.gitignore"
 cat <<EOT > "${name}port/.gitignore"
 log/
 install/
-${name}/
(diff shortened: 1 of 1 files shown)


bin/lib/zopen-generate
```diff
@@ -172,6 +172,7 @@ touch "${name}port/.gitignore" && chtag -tc 819 "${name}port/.gitignore"
 cat <<EOT > "${name}port/.gitignore"
 log/
 install/
+${name}*/
```
MikeFultonDev, line 175: isn't this a fairly broad ignore entry? This could accidentally pull in directories people may develop, e.g. gittests (say) ... I am expecting this is for the particular tarball and git directories created from the extract?

IgorTodorovskiIBM: It could probably be improved to get the basename from the tar and git urls and add both using something similar to this code:
```
filename=$(basename "$url") # get basename
name="${filename%.*}" # strip out the extension
```

v1gnesh: I think the directory `gitpatches` will get caught too?

gitport

IgorTodorovskiIBM: @v1gnesh , is this PR ready or is there more to be done?

v1gnesh: We don't have a way to allow dirnames like `gitpatches`.
Per this PR, `gitport` will add `git*/` to `.gitignore`, and that'll add gitpatches/ to ignore as well.

PR #314, opened 2023-06-01
Add an option for force apply the patches, useful when moving up to new versions

IgorTodorovskiIBM:
* When moving up to a newer versions and when the patches don't cleanly apply, it's useful to force apply all of the patches and then to manually patch in the rejected changes. This change enables this. The rejected changes will be added to a corresponding `.rej` file
* Use `zopen build -v --forcepatchapply`


MikeFultonDev (approved):
lgtm

PR #315, opened 2023-06-01
simple tool to print out brew and z/os tool versions

MikeFultonDev:
This tool is a little hokey - but it's a good 'one off' tool to see what packages we have are lower level than what brew has

Lots of room for improvement.

### DIFF
--- a/bin/zopen-versions
+++ b/bin/zopen-versions
@@ -1,0 +1,65 @@
+#!/bin/env bash
+
+#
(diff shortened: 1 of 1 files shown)


IgorTodorovskiIBM (approved):

IgorTodorovskiIBM:
It would be cool if we can integrate this into the docs or in the progress page (maybe via https://github.com/ZOSOpenTools/meta/blob/main/tools/getbinaries.py).

PR #330, opened 2023-06-12
fixed #076

### LINKED ISSUE #76: zopen-setup: if no parameters are passed, create directory under $HOME/zopen
In addition, create the directory if it doesn't exist.

sabitor:
The way how `zopen-setup` works now is as follows:

1. `zopen-setup` can be invoked without specifiying a root-directory for the zopen tools. In this case, the setup process creates a `$HOME/zopen` directory and installs the zopen tools there.
2. `zopen-setup` can be invoked by specifiying a root-directory for the zopen tools to be installed. 
The root directory doesn't have to exist. If it's there it will be taken, if not it is going to be created. 
Thereby the root directory can be either as an absolute path or as a relative path, for example: 
\- absolute root-directory: `zopen-setup /home/user/projects/dev/zopen`
\- relative root-directory: `zopen-setup projects/test/zopen` 
In either case, the missing parts of the root-directory (if there are any) are created as part of the setup process (if the permissions allow to do so).

### DIFF
--- a/docs/Guides/QuickStart.md
+++ b/docs/Guides/QuickStart.md
@@ -35,7 +35,7 @@ It will be much easier to work with USS using ssh, scp, and sftp.
 ### Running zopen-setup
 
 - `cd $HOME`
-- `./zopen-setup  <file system>` # file system is the location where the zopen tools will be installed
+- `./zopen-setup` # The zopen tools will be installed into `$HOME/zopen`. If a path parameter is specified, the zopen tools will be installed there. The specified path can either be an absolute path or a relative path. If the specified path does not exist, it will be created as part of the setup process.
 
 ### Set up the zopen tools
 
--- a/tools/src/zopensetupmain.c
+++ b/tools/src/zopensetupmain.c
@@ -3,6 +3,7 @@
 #include <stdlib.h>
 #include <stdio.h>
 #include <string.h>
+#include <unistd.h>
 
 #include "download.h"
 #include "zopenio.h"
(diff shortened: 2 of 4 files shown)


tools/src/zopensetupmain.c
```diff
@@ -35,7 +36,7 @@ static void syntax(const char* pgm, const char** bootpkg) {
     pgm = &base[1];
   }
   fprintf(stderr, "%s : Install the z/OS Open Source Tools 'starter' environment by downloading from %s\n"
-                  "Syntax: %s [-vq] <root>\n"
+                  "Syntax: %s [-vq] [<root>]\n"
```
IgorTodorovskiIBM, line 39:
```suggestion
                  "Syntax: %s [-vq] [root directory]\n"
```
<> usually means required argument, so I would suggestion removing it

tools/src/zopensetupmain.c
```diff
@@ -35,7 +36,7 @@ static void syntax(const char* pgm, const char** bootpkg) {
...
                   "  <root> is the root directory where:\n"
```
IgorTodorovskiIBM, line 40:
```suggestion
                  "  [root] is optional root directory where:\n"
```

docs/Guides/QuickStart.md
```diff
@@ -35,7 +35,7 @@ It will be much easier to work with USS using ssh, scp, and sftp.
 ### Running zopen-setup
 
 - `cd $HOME`
-- `./zopen-setup  <file system>` # file system is the location where the zopen tools will be installed
+- `./zopen-setup` # The zopen tools will be installed into `$HOME/zopen`. If a path parameter is specified, the zopen tools will be installed there. The specified path can either be an absolute path or a relative path. If the specified path does not exist, it will be created as part of the setup process.
```
IgorTodorovskiIBM, line 38:
```suggestion
- `./zopen-setup [optional path]` # The zopen tools will be installed into `$HOME/zopen`. If a path parameter is specified, the zopen tools will be installed there. The specified path can either be an absolute path or a relative path. If the specified path does not exist, it will be created as part of the setup process.
```
Square brackets usually denote optional arguments. Wonder if that is clearer or not?

IgorTodorovskiIBM (approved):
Minor comments, looks good otherwise. Thank you!

sabitor:
With the release of the new zopen package manager, zopen-setup is no longer used. 
Thus, this PR can be closed.

PR #334, opened 2023-06-16
Modify zopen-build to generate the .depsenv rather than zopen-install

IgorTodorovskiIBM:
Reasons for doing this:
* sourcing of .depsenv is currently appended via the zopen-install, but this is an issue if we want to use the variables or tools from the dependencies in the setup.sh script.
The current flow is: (program is sourced)->(setup.sh is run)->(dependencies are sourced)
The new flow will be (program is sourced)->(dependencies are sourced)->(setup.sh is run)
* It will also allow individual .pax.Z downloads to potentially pick up dependencies

### DIFF
--- a/bin/lib/zopen-build
+++ b/bin/lib/zopen-build
@@ -1239,6 +1239,42 @@ touch \".installed\"
 rm \".installing\"
 zz
 
+if [ ! -z "$ZOPEN_RUNTIME_DEPS" ]; then
+cat <<ZZ >> "${ZOPEN_INSTALL_DIR}/.env"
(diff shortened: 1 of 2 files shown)


MikeFultonDev (commented):
I am not quite understanding. Can you provide an example where this solves a problem?

IgorTodorovskiIBM:
> I am not quite understanding. Can you provide an example where this solves a problem?

This resolves an issue with gpg, where it depends on pinentry and needs to know the path of pinentry-tty at setup.sh time.

MikeFultonDev (approved):
Lgtm

PR #338, opened 2023-06-20
Add more attempts when update-cacert fails as it sometimes times out on certain systems

IgorTodorovskiIBM:
* Plus enable verbose output for certain Makefile that check for V=1 passed to Make (tig and emacs)
* Plus update the cacert.pem


MikeFultonDev (approved):
LGTM

PR #342, opened 2023-06-23
Fix zopen build errors

IgorTodorovskiIBM:


MikeFultonDev (approved):
i hit the same issues but hadn't promoted yet. thank you!

PR #348, opened 2023-06-27
Install zopen into the usual path rather than the current directory

IgorTodorovskiIBM:
Use similar code to the dependency detection where it installs into the first path of the zopen search path.


MikeFultonDev (approved):
lgtm

PR #369, opened 2023-08-14
Fix json cache to report the correct tests statistics

IgorTodorovskiIBM:
Previously, it was reporting total tests and success as -1 as seen here:
https://github.com/ZOSOpenTools/meta/blob/main/docs/api/zopen_releases.json


MikeFultonDev (approved):
LGTM

PR #376, opened 2023-09-01
Merge Russell's zopen fork into main

IgorTodorovskiIBM:
Merging https://github.com/DevonianTeuchter/meta


README.md
```diff
@@ -1,90 +1,100 @@
...
+zopen install which
+zopen list --installed
+which which
+zopen upgrade
+```
 
-But - there is a _transitive closure_ problem to address with porting a software package, namely understanding what packages are pre-requisites
-for the software package you want to port. As an example, for zsh, we see the following:
+## Important usage notes
+- On first run, ```zopen init``` will copy this forked version of meta into the "package" area of zopen (ie. where packages are expanded and accessed from).  It will also be pinned to this release to prevent any possible updates from the "real" meta package. Removing the .pinned file from the meta-dt directory will allow for the main meta port to be installed however this will cause incompatabilities if run.
```
MikeFultonDev, line 38: Is this still right now that it isn’t a fork?

README.md
```diff
@@ -1,90 +1,100 @@
...
+- Removing zopen and the z/OS Open Tools once installed involves: uninstalling all installed packages; removing any copies of ```.zopen-config```; removing the configured zopen root directory (by default ```/usr/local/zopen``` but is set during installation); and then running a command to find any final orphaned symlinks on the system, such as: ```/bin/find $ZOPEN_ROOTFS -type l -exec test ! -e {} \; -print```  where $ZOPEN_ROOTFS is '/' [replace `-print` with `rm -rf` to actually remove symlinks, the example command should only list what was found - care should be taken when removing any files with sysadmin authority to prevent removing critial files!]
 
-GNU make requires GNU m4, automake, autoconf, Perl, and a C compiler that is gcc compatible.
 
-## Order to Build _from scratch_
+## Problem resolution
+If the meta package does get updated to a non-dt version, then running ```./zopen alt meta -s``` from the meta-dt/bin directory within the z/OS Open Tools root filesystem should allow selection of the dt-forked meta; running any other non-forked zopen commands might break the installation or cause unexpected behaviour!
```
MikeFultonDev, line 48: Is this now old since ‘dt’ is not a thing ?

README.md
```diff
@@ -1,90 +1,100 @@
...
-| [IBM XL C/C++ V2.4.1](https://www-40.ibm.com/servers/resourcelink/svc00100.nsf/pages/xlCC++V241ForZOsV24) | IBM [^ibm] | [ibm.com web download](https://www.ibm.com/marketing/iwm/iwm/web/dispatcher.do?source=swg-zosxlcc) |
-| [git](https://git.kernel.org/pub/scm/git/git.git/) | [LGPL V2.1](https://git.kernel.org/pub/scm/git/git.git/tree/LGPL-2.1) | [z/OS Open Tools release](https://github.com/ZOSOpenTools/gitport/releases/tag/boot)  |
-| [curl](https://github.com/curl/curl) | [curl-license](https://github.com/curl/curl/blob/master/COPYING) | [z/OS Open Tools release](https://github.com/ZOSOpenTools/curlport/releases/tag/boot) |
-| [gunzip](https://www.gnu.org/software/gzip/) | [GPL V3](https://www.gnu.org/licenses/gpl-3.0.html) | [z/OS Open Tools release](https://github.com/ZOSOpenTools/unzipport/releases/tag/boot) |
+>zopen install <package>...
 
-[^ibm]: a no-charge add-on feature for clients that have enabled the XL C/C++ compiler (an optionally priced feature) on z/OS
+Used to install a z/OS Open Tools package.  By default, the latest version is installed although options are available to install specific versions, tags or to pick from a selection [see ```--help``` for more]
```
MikeFultonDev, line 62: I thought stable was the default not latest?

MikeFultonDev (commented):
First comments. Not done.

README.md
IgorTodorovskiIBM: Updated

README.md
IgorTodorovskiIBM: Updated

Issue #377, opened 2023-09-04
Doing 'zopen build' twice gives error about existing directory on second build


MikeFultonDev:
This is for 'zopen build' of makeport but I expect it would be the same for others...

![image](https://github.com/ZOSOpenTools/meta/assets/61100689/96f308fd-9893-4269-bb87-f37d45f68f83)

MikeFultonDev:
confirmed this is a 'general' bug - see the same thing with vim:
![image](https://github.com/ZOSOpenTools/meta/assets/61100689/36ad65d1-91f1-46f7-8e65-fc0a2f46d7a3)

IgorTodorovskiIBM:
Hmm, I just tried it with `zotsample`, and I didn't encounter this issue. But `zotsample` may not be typical. I'll try vim next

IgorTodorovskiIBM:
I can reproduce with vim:
```
Generating symlink from /jenkins/zopen/usr/local/zopen/vim to /jenkins/zopen/usr/local/zopen/vim-master
VERBOSE: 'if [ -h /jenkins/zopen/usr/local/zopen/vim ]; then rm /jenkins/zopen/usr/local/zopen/vim; fi && ln -s /jenkins/zopen/usr/local/zopen/vim-master /jenkins/zopen/usr/local/zopen/vim'
ln: FSUM6245 link to target "/jenkins/zopen/usr/local/zopen/vim/vim-master" failed: EDC5117I File exists.
***ERROR: Link command failed
bash-4.3$
```

PR #383, opened 2023-09-05
Address zopen build's install location

IgorTodorovskiIBM:
This matches the expected install location of the new zopen installer. This should also address https://github.com/ZOSOpenTools/meta/issues/377


DevonianTeuchter (approved):
LGTM.

PR #384, opened 2023-09-05
Remove dot from zopen-config, so it's not hidden

IgorTodorovskiIBM:


DevonianTeuchter (approved):
LGTM  [perhaps we should have extracted the file name to a constant somewhere initiallyto make this easier :-)  ]

PR #385, opened 2023-09-05
Add a suggestion if zopen-config is not loaded

IgorTodorovskiIBM:


DevonianTeuchter (changes requested):
After the logic from 683-687, there needs to be a printError at line 687 otherwise processing will continue, even if the $ZOPEN_CA could not be read.
Rather than the relative path, would it make sense to use an absolute path for clarity/length?

MikeFultonDev (approved):
LGTM

DevonianTeuchter:
LGTM

DevonianTeuchter (approved):

PR #390, opened 2023-09-06
Optimize .env and .zopen-config processing

IgorTodorovskiIBM:
Since the new zopen symlinks all binaries, libraries, and man pages under a common bin, lib, share location, we no longer need to set the PATH, LIBPATH and MANPATH environment variables for each tool.

However, if someone downloads the tool manually, then we should still retain the existing .env logic. So this change allows for both cases to work.

The idea is that we would decouple the required tool environment variables away from the .env into a .appenv, which zopen-config would source. This would greatly speed up the .zopen-config processing.

This would require a new build of all tools that have zopen_append_to_env logic.

There's currently a fallback in place, such that if .appenv doesn't exist, it will source .env.

If there's any environment variables that are used within the application process, they should use `zopen_append_to_zoslib_env` (as in the case of Git):
```
zopen_append_to_zoslib_env() {
cat <<EOF
GIT_TEMPLATE_DIR|set|PROJECT_ROOT/share/git-core/templates
GIT_EXEC_PATH|set|PROJECT_ROOT/libexec/git-core
GIT_SSL_CAINFO|set|PROJECT_ROOT/cacert.pem
EOF
}
```

But variables that are needed by other tools or processes should go into `zopen_append_to_env` (we might need to change the name) as in the case of autoconf:
```
zopen_append_to_env()
{
cat <<EOF
export autom4te_perllibdir=\${PWD}/share/autoconf
export AC_MACRODIR=\${PWD}/share/autoconf


#To set M4 variable path
if [ -z "${M4_HOME}"  ]; then
    echo "Ensure M4_HOME is set and re-source the .env"
else
    export M4="\$M4_HOME/bin/m4"
fi
EOF
}
```

### DIFF
--- a/bin/lib/zopen-build
+++ b/bin/lib/zopen-build
@@ -1268,9 +1268,12 @@ zz
   if command -V "${ZOPEN_APPEND_TO_ENV_CODE}" >/dev/null 2>&1; then
     printVerbose "Appending additional environment variables..."
     append_to_env="$(${ZOPEN_APPEND_TO_ENV_CODE})"
-    echo "$append_to_env" > "${ZOPEN_INSTALL_DIR}/.appenv"
-    echo ". ./.appenv" >> "${ZOPEN_INSTALL_DIR}/.env"
+  else
+    append_to_env="# nothing to do"
   fi
+
+  echo "$append_to_env" > "${ZOPEN_INSTALL_DIR}/.appenv"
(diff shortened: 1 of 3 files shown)


MikeFultonDev (approved):
Lgtm

DevonianTeuchter (approved):
LGTM.

PR #399, opened 2023-09-08
Change the install location so as not to conflict with the default zopen install location. 

IgorTodorovskiIBM:
* Installs locally built packages to ${ZOPEN_PKGINSTALL}/dev/
* Also adds it to the dependency search path
* This will enable tools that have build dependencies on other tools/libraries to pick up the local dev builds


DevonianTeuchter (approved):
Seems ok to me

Issue #402, opened 2023-09-11
Ctrl-C during zopen-config processing exits the current session


IgorTodorovskiIBM:
Steps to reproduce:
```
. /path/to/zopen-config # press ctrl-c during processing
```
It will exit the current session

Issue #400, opened 2023-09-09
zopen list --upgradeable shows things that are not upgradeable


MikeFultonDev:
![image](https://github.com/ZOSOpenTools/meta/assets/61100689/16bef223-2202-4b87-a3c9-bfb37286305f)

IgorTodorovskiIBM:
The contents of .releaseinfo and .version appear to be the same?
```
[ITODORO@ZOSCAN2B ~/projects/zoslibport]$ cat ~/zopen-official/usr/local/zopen/make/make/.releaseinfo
4.4.1.20230705_115215
[ITODORO@ZOSCAN2B ~/projects/zoslibport]$ cat ~/zopen-official/usr/local/zopen/make/make/.version
4.4.1.20230705_115215
```

IgorTodorovskiIBM:
I copied the logic from zopen upgrade to zopen query: https://github.com/ZOSOpenTools/meta/pull/403

PR #403, opened 2023-09-11
Fix zopen list --upgradeable + zopen-config fix

IgorTodorovskiIBM:
Fixes https://github.com/ZOSOpenTools/meta/issues/400
and fixes https://github.com/ZOSOpenTools/meta/issues/402


DevonianTeuchter (approved):
LGTM - (For reference: the versioning/release was legacy to handle packages that didn't quite have a consistent set of release/version files - as packages etc mature, things are more consistent!

PR #407, opened 2023-09-15
Pass thru custom curl parameters

DevonianTeuchter:
Allow for users to add custom parameters to the curl command to allow options such as proxies.
Also remove some unused code now the JSON cache is implemented

### DIFF
--- a/bin/lib/zopen-build
+++ b/bin/lib/zopen-build
@@ -829,7 +829,7 @@ downloadTarBall()
   else
     freshBuild=true
     if ${verbose}; then
-      printVerbose "$curlCmd -L -o ${tarballz} ${ZOPEN_URL}"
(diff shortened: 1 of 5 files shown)


bin/lib/zopen-build
```diff
@@ -829,7 +829,7 @@ downloadTarBall()
   else
     freshBuild=true
     if ${verbose}; then
-      printVerbose "curl -L -o ${tarballz} ${ZOPEN_URL}"
+      printVerbose "$curlCmd -L -o ${tarballz} ${ZOPEN_URL}"
```
IgorTodorovskiIBM, line 832: Would this work for a function?

IgorTodorovskiIBM (approved):
Look good to me!

bin/lib/zopen-build
DevonianTeuchter: will just print the blank string "".  Added new commit

PR #411, opened 2023-09-15
Use  <rootfs>usr/local/  as  base rather than

DevonianTeuchter:
Make default to generate symlink map in <rootfs>/usr/local/... rather than <rootfs>/usr/...


MikeFultonDev (approved):
Lgtm

PR #412, opened 2023-09-16
Add GH Actions action.yml

v1gnesh:
Instead of creating custom actions in each repo, the action can live in `meta`, and can be reference in repo-specific workflow files like so - https://github.com/ZOSOpenTools/duckdbport/blob/main/.github/workflows/bump.yml

```yaml
name: 'Automatic version updates'

on:
  schedule:
    # minute hour dom month dow (UTC)
    - cron: '00 15 * * *'
  # enable manual trigger of version updates
  workflow_dispatch:
jobs:
  bump:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@master
      - uses: zosopentools/duckdbport/action@main
        env:
          GITHUB_TOKEN: ${{ secrets.BUMP_TOKEN }}
```

Reference to an automatically generated PR: https://github.com/ZOSOpenTools/duckdbport/pull/2

Additionally, it would be great if an org-wide secret called `BUMP_TOKEN` can be created and populated with a Personal Access Token that has the `repo` check boxes checked for repo read access.
Also would be best if the PAT belonged to the `ZOSOpenTools` ID.

Feel free to change `user.name`, `user.email` or anything else in the `action.yml` as required.

The action finishes within 20s, which is great.
It's currently setup to run every day at 1500 UTC.

**EDIT:**
Can also add a neat little badge to the participating repos like so:
image

### DIFF
--- a/actions/action.yml
+++ b/actions/action.yml
@@ -1,0 +1,42 @@
+name: 'Bump versions'
+description: 'Use bump to create version update pull requests'
+author: 'ZOSOpenTools'
+branding:
+  icon: 'chevrons-up'
+  color: 'green'
+runs:
+  using: 'docker'
+  image: 'docker://mwader/bump'
+inputs:
+  bumpfile:
+    description: 'Bumpfile to read'
+    required: false
+    default: 'Bumpfile'
+  bump_files:
+    description: 'Files with embedded configuration or versions to update'
+    required: false
+    default: 'buildenv'
+  title_template:
+    description: 'Commit and pull request title template'
+    required: false
+    default: 'Update {{.Name}} to {{.Latest}} from {{join .Current ", "}}'
+  commit_body_template:
+    description: 'Commit body template'
+    required: false
+    default: '{{range .Messages}}{{.}}{{"\n\n"}}{{end}}{{range .Links}}{{.Title}} {{.URL}}{{"\n"}}{{end}}'
+  pr_body_template:
+    description: 'Pull request body template'
+    required: false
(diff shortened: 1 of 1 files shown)


MikeFultonDev (commented):
I have to admit I don’t know how the larger flow of this works. I think you talked about this on a guild call though right? Can you point me to that so I can read up a bit?

v1gnesh:
Sure, [here](https://github.com/ZOSOpenTools/gitport/pull/74)'s an example of what bump does (in general).
What it does in the Action can be seen from the workflow log [here](https://github.com/ZOSOpenTools/duckdbport/actions/runs/6206882635/job/16851558756).

As far as the GH Action goes, it's like this:

1. Create a workflow called `<some>port/.github/workflows/bump.yml`. This workflow points to a schedule (cron timer), GITHUB_TOKEN, the type of Action runner, and the `Action`s to use.
2. Currently, the Action used is `duckdbport/action/action.yml`, and in it, in the `inputs` section, in `bump_files`, we tell bump to look for embedded commands in the `buildenv` file (see the top of `buildenv`).
Embedded bump commands look like this:
```shell
# bump: git /GIT_VERSION="(.*)"/ https://github.com/git/git.git|*
```
For projects that create releases & tags, it's easy.
But for GNU projects and such, regex is needed to parse content on a specific HTML page. 
All of that is done, and embedded commands are now ready for all projects; just need to push them to each repo (with myself & Igor).
Once they're in, we can turn on Actions for all repos.

3. This PR is to move that Action file to meta, so that there's no duplication of the same file across all repos.
However, `.github/workflows/bump.yml` may be required everywhere... unless org-wide workflows can be setup (have to look it up).

MikeFultonDev (approved):
Thanks for the description !

IgorTodorovskiIBM (approved):
LGTM

PR #415, opened 2023-09-17
zopen build install changes + fix for replacing hardcoded paths in package contents

IgorTodorovskiIBM:
After discussing with @MikeFultonDev, we decided to change the zopen build install location back to `${ZOPEN_PKGINSTALL}/<pkg>/<pkg>` instead of `${ZOPEN_PKGINSTALL}/dev/<pkg>/<pkg>`. This is so that we can leverage the `zopen alt` logic to optionally set the local build as the current release.

Here's an example of zoslib. 

Before zopen build:
```
`: zoslib-zopen.20230914_230230.zos  <-  current
```

After zopen build:
```
1: zoslib-new_file_env  <-  current
2: zoslib-zopen.20230914_230230.zos
```
* Additionally, we _probably_ don’t need to change the ZOPEN_INSTALL_DIR naming logic because it doesn’t contain a timestamp, which means it will never conflict with official install paths.

If the user does not want to set their local build as active, then they can add the option:
* `—no-set-active` will not affect the active version

This also fixes the issue where hardcoded paths to the root directory are not replaced. Additionally, since the paths contains a timestamp, additional logic has been added to strip them away.
**Note:** https://github.com/ZOSOpenTools/meta/compare/zopen-build-install-changes?expand=1#diff-18d8a0f3b6e2204467238e541740a1e6b0092f69dcd8f83bbf3d94627f172e69R1202 - to ensure the replace hardcoded paths logic works, the path suffix is removed. For example, zoslib-zopen.20230914_230230.zos is modified to zoslib.

### DIFF
--- a/bin/lib/zopen-build
+++ b/bin/lib/zopen-build
@@ -240,7 +240,6 @@ printSyntax()
   echo "  -g|--get-source: get the source and apply patch without building"
   echo "  -gp|--generate-pax: generate a pax.Z file based on the install contents"
   echo "  --forcepatchapply: force apply the patches, where rejected patches are placed into a corresponding file of the same name, with the .rej extension"
-  echo "  -nosym|--nosymlink: do not generate a symlink from the project name to \${ZOPEN_INSTALL_DIR}"
   echo "  --no-set-active: do not change the pinned version"  >&2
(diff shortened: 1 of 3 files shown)


MikeFultonDev (approved):
An elegant solution I think!

Issue #420, opened 2023-09-19
zopen install --all is broken


IgorTodorovskiIBM:
@ccw-1 reported that zopen install --all is broken

```
zopen install --all
Enter 'all' to confirm full installation [takes a long time so be sure!]:
all
- Querying remote repo for latest package information
Installing package:
***ERROR: Could not get all releases for
```

PR #422, opened 2023-09-19
Fix zopen install --all

IgorTodorovskiIBM:
Fixes https://github.com/ZOSOpenTools/meta/issues/420

Also Implicitly set alt when -d option is specified


MikeFultonDev (approved):
LGTM

PR #428, opened 2023-09-21
426 zopen clean  a groff didnt do what i expected

### LINKED ISSUE #426: zopen clean -a groff didn't do what i expected
I have 3 packages in groff:

I was expecting zopen clean -a groff to clean up the ones I am not using. Here's what I got:

```
FULTONM@ZOSCAN2B bash ~> zopen clean -?
zopen clean is a utility for z/OS Open Tools to switch package versions for currently installed packages
Syntax: zopen-clean [<option>]* [<package]
  where <option> may be one or more of:
      -u : --unused: removes unused package versions
      -d : --dangling: removes dangling symlinks
      -c : --cache: cleans package cache
      -a : --all:  equivalent to -u -d -c
  -v: run in verbose mode.
TODO and <package> is a package name.
FULTONM@ZOSCAN2B bash ~> zopen clean -a groff
- Dangling link removal complete
- Cache at '/home/fultonm/zopen/var/cache/zopen' cleaned
FULTONM@ZOSCAN2B bash ~> #exit
FULTONM@ZOSCAN2B bash ~> zopen alt groff
1: groff-1.22.4.20230622_204353.zos
2: groff-1.22.4.20230918_154912.zos
3: groff-1.22.4.20230920_020202.zos  <-  current
```

Also - the help doesn't say what the default is if you don't specify an option...

@IgorTodorovskiIBM @DevonianTeuchter would appreciate your input

DevonianTeuchter:
Should resolve #426 
Removed the -a option as it was confusing [how does -u work for a -a when needs a parameter and the cache/symlinks are zopen-system-wide and not package-specific so having them combinable doesn't make sense.  
Improved the help with examples
Improved package detection logic

### DIFF
--- a/bin/lib/common.inc
+++ b/bin/lib/common.inc
@@ -15,6 +15,15 @@ cleanupFunction() {
 }
 trap "cleanupFunction" EXIT INT TERM QUIT HUP
 
+isPackageActive(){
+  pkg="$1"
+  printDebug "Checking if '$pkg' is installed and active"
+  installedPackage=$(cd "$ZOPEN_PKGINSTALL" && zosfind . -name ".active" | grep "/$pkg/")
+  cmdrc=$?
+  # Return 1 for true/Package is Active, 0 for false/Package is not active
+  [ $cmdrc -eq 0 ] && return 1 || return 0
+}
+
 curlCmd(){
   # Take the list of parameters and concat them with  
   # any custom parameters the user requires in ZOPEN_CURL_PARAMS
--- a/bin/lib/zopen-alt
+++ b/bin/lib/zopen-alt
@@ -5,33 +5,97 @@ export utildir="$( cd "$(dirname "$0")" >/dev/null 2>&1 && pwd -P )"
 
 . "${utildir}/common.inc"
 
-printSyntax() 
-{
-  args=$*
-  echo "zopen alt is a utility for z/OS Open Tools to switch package versions for currently installed packages" >&2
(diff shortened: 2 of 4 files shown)


IgorTodorovskiIBM:
Nice! 

I see you removed code relevant to "Installing and pinning currently running version of meta". If I want to install my "experimental" version from my local meta workspace, can I still do that? If not, wonder if we can provide that as an option

DevonianTeuchter:
Yes, you can still use your experimental option [it's what I do] 
If you have your code in a "meta-my-experiments-dev" directory in `<root>/<installpath>/meta/`  for example, `/zopen/usr/local/zopen/meta/meta-my-experiments-dev`, then you can should be able to switch to that version using `zopen alt --set meta meta-my-experiments-dev`  (assuming you have the latest zopen-alt changes) or `zopen alt --set meta` and select your listed version (if you have the pre-alt changes!).  Would be a good independent test of that capability...

MikeFultonDev (approved):
thought i already approved this. great enhancement.

DevonianTeuchter:
> 
> Ah ok, but I would need to copy it to that directory first? (If I'm not already working off of that directory)?
True
It might be possible to symlink your work directory into the .../meta/  directory so nothing needs moving and might just work... *"might"* being the operative word as there are some searches for symlinks.
Or hardlink the files into a new directory tree under  .../meta so the directory is standard but the files reside in your work directory.
Haven't tested either of those scenarios tbh the second should just work, the first might.

IgorTodorovskiIBM (approved):
LGTM

PR #429, opened 2023-09-21
Update action.yml

v1gnesh:
Can use `files:` instead of `bump_files:` now.
Ref: https://github.com/wader/bump/pull/120


IgorTodorovskiIBM (approved):
LGTM

PR #432, opened 2023-09-21
Fix libpath and manpath to reflect symlink paths

IgorTodorovskiIBM:
* Keep the only ones as well so as not to cause error to existing installations.


MikeFultonDev (approved):
LGTM

PR #434, opened 2023-09-21
add version and have zopen know about itself to eliminate need for .env

MikeFultonDev:
@IgorTodorovskiIBM and @DevonianTeuchter please review. 
Looking to simplify steps to get started with zopen

I also bumped the version to 1.0.1 so there isn't any confusion about what is 'better' since we have an old 1.0.0

### DIFF
--- a/bin/lib/zopen-version
+++ b/bin/lib/zopen-version
@@ -1,0 +1,3 @@
+#!/bin/sh
+
+echo "zopen 1.0.1"
--- a/bin/lib/zopen-version
+++ b/bin/lib/zopen-version
@@ -1,3 +1,3 @@
 #!/bin/sh
 
-echo "zopen 1.0.1"
+echo "zopen 0.7.0"
--- a/.env
+++ b/.env
@@ -1,4 +1,10 @@
 #
-# Setup no longer required. 'zopen' finds itself
+# Setup no longer required if you are just 'using' tools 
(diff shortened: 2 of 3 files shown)


bin/lib/zopen-version
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+echo "zopen 1.0.1"
```
IgorTodorovskiIBM, line 3: Wonder if we should have a seperate ZOPEN_VERSION file that can be referenced here and potentially in other places or we expect everything will go through zopen-version?

IgorTodorovskiIBM: Also, 1.0.1 seems like a big jump or 0.6.3. Should we go with 0.7.0?

MikeFultonDev: i was thinking all zopen tools can just call zopen-version to print out the version so we don't have to worry about keeping things in sync

1.0.1 is not much of a jump from 1.0.0 (which we did before the 0.6.x)

happy to make it 0.6.4 if we delete the 1.0.0 pax file

MikeFultonDev: added code back in for .env with comment to indicate it isn't required for consumers
changed bump to 0.7.0 instead of 1.0.1

IgorTodorovskiIBM (approved):
LGTM

PR #449, opened 2023-09-25
remove pinning "real meta" note

covener:
IIUC now that zopen-dt is rolled in, there is no default pin and no "real" meta to dodge.


MikeFultonDev (approved):
Good catch. My understanding is there are no restrictions with meta being upgradeable now.
@DevonianTeuchter please comment if otherwise.

PR #453, opened 2023-09-25
rename version file from .version to zopen-version 

MikeFultonDev:
so symlink will get created in metaport


IgorTodorovskiIBM (approved):
LGTM

PR #455, opened 2023-09-25
Updates to CI build step to upgrade meta, clean cache

IgorTodorovskiIBM:
* Automatically upgrade meta on every build
* Clean the cache after successful builds


MikeFultonDev (approved):
nice.

PR #459, opened 2023-09-26
Code restructuring and bug fix for umask

MikeFultonDev:
- Changed umask to be umask 022 so that code won't be marked group rw
- Remove bold from zopen help so that it can be processed by help2man
- Move bin/lib/common.inc up to include/common.sh
- Move bin/data to data

- Moved all zopen-* tools into the common 'bin' directory and cleaned up internal links

I ran the various tools:
- zopen install
- zopen build zotsampleport
- zopen-help2man
- zopen remove
- zopen version
- zopen init


IgorTodorovskiIBM (approved):
LGTM

PR #460, opened 2023-09-26
Convert properties.json to metadata.json and add more fields

IgorTodorovskiIBM:
This PR will add metadata to the pax. I will also add it outside of the pax file so that we can quickly process it when generating the release cache.

What other fields should we add?
* [x] name
* [x] version 
* [x] category or group?
* [x] repository?
* [x] install size
* [x] test status
* [x] build line
* [x] runtime deps    
* [x] build deps
* [x] commit sha
* [x] md5 checksum?
* [x] pax size?
* [x] license 

Future:  

* gpg signature? 
* BOM related materials? license? 
* excludes to exclude unnecessary dirs


IgorTodorovskiIBM:
@DevonianTeuchter curious what your thoughts are here?

DevonianTeuchter:
checksum - yes (do we need to specify the algorithm?)
signature - is this the GPG sig?  Would be good to start implementing those (but would we ship a bootstrap gpg? )

license - would be useful along with...
BOM related materials -  could add a "zopen audit" capability

repository - not sure this would add much??
issue link - probably not
install size - yes
pax size - maybe

DevonianTeuchter:
Also thinking... 
if the pax was to contain anything extra that doesn't need to be applied to a system; for example, the "meta" package itself ships a lot of directories and tools that might not really be relevant for a running system - like "tools" or "cicd".  There could be an "excludes" property (or conversely an "includes" property). 
Maybe, commit history so that if a particular version broke something, then that would show in the pax's commit history (allowing a level to be "--select"'d for install prior to the breaking commit)  - and something that could be displayed in an audit command also.

v1gnesh:
Hi both, check this out - https://github.com/orgs/ZOSOpenTools/discussions/212#discussioncomment-7129712

IgorTodorovskiIBM:
Adding category or group as discussed in discord

IgorTodorovskiIBM:
I have updated the PR with a few more metadata fields, including size and category

IgorTodorovskiIBM:
An example for ctags:
```
{
  "product": {
    "name": "ctags",
    "version": "5.8.0.0",
    "release": "20231027_155500",
    "summary": "ctags on z/OS",
    "repo": "git@github.com:ZOSOpenTools/ctagsport.git",
    "categories": "editor",
    "total_tests": "1",
    "total_success": "1",
    "buildline": "STABLE",
    "size": "1886",
    "runtimedeps": "",
    "builddeps": "comp_xlclang curl git gzip jq make sed tar zoslib",
    "commitsha": "b5eb65e86ab492c733cae2b8574731f747494b28",
    "pax": "ctags-5.8.20231027_155500.zos.pax.Z",
    "pax_size": "784",
    "md5": "de6f592ec3423c2fc693baec59f12cf7"
  }
}
```

MikeFultonDev:
Can you compare / contrast with information other package managers have? In particular there was one you thought we may want to move to so consistency with it would be good. 
For runtime and build should we be listing minimum version required or what we built/tested with?

MikeFultonDev:
> Can you compare / contrast with information other package managers have? In particular there was one you thought we may want to move to so consistency with it would be good. For runtime and build should we be listing minimum version required or what we built/tested with?

@IgorTodorovskiIBM in particular, while we don't have any specific minimum / maximum versions for build/runtime dependencies, we should design in what that would look like. Also we should have a version in here somewhere so that when we need to change, we can tell what version of metadata we are looking at.

IgorTodorovskiIBM:
Latest update include version_scheme, licenses, and improved formatting

Here's an updated look at the metadata.json:
```json
{
  "version_scheme": "0.1",
  "product": {
    "name": "autoconf",
    "version": "2.71.0.0",
    "release": "20231103_121123",
    "summary": "autoconf on z/OS",
    "repo": "https://github.com/ZOSOpenTools/autoconfport",
    "license": "https://github.com/ZOSOpenTools/autoconfport/blob/main/patches/LICENSE",
    "zopen_license": "https://github.com/ZOSOpenTools/autoconfport/blob/main/LICENSE",
    "categories": "",
    "buildline": "STABLE",
    "test_status": {
      "total_tests": "",
      "total_success": "-1"
    },
    "size": "7852",
    "build_dependencies": "autoconf automake bash comp_xlclang curl flex git gzip jq m4 make perl sed tar texinfo zlib zoslib",
    "commitsha": "eefb4418c29453d19d5ef59fcbb4fd3e5433117c",
    "runtime_dependencies": [
      {
        "name": "m4",
        "versionRange": null
      },
      {
        "name": "perl",
        "versionRange": null
      }
    ],
    "pax": "autoconf-2.71.20231103_121123.zos.pax.Z",
    "pax_size": "2800",
    "md5": "ec23d964e5ca4e0c2757c6dd430b755a"
  }
}
```
versionRange is null for now, but we can use the same logic as we implemented for build dependencies, which allows for >, >=, and other operators. It seems that brew provides just a list of dependencies and specific version: https://formulae.brew.sh/api/formula/git.json

I did look at brew and rpm spec. There are some interesting ones that we could add at some point, like "caveat" in brew which could be useful for tools like sudo that require some steps to be run with superuser authority.

MikeFultonDev (approved):
LGTM

DevonianTeuchter (approved):
LGTM and opens up a lot of opportunities!

PR #461, opened 2023-09-26
clean up options processing

MikeFultonDev:
Change code so that it scans all the input args and partitions them into subopts or subcmd.
Then process the options --help and --version up front
Then call the subcmd


IgorTodorovskiIBM (approved):
nice! LGTM

PR #464, opened 2023-09-26
restrict tools that perform an update to writable zopen tool chains

### LINKED ISSUE #463: restrict zopen- commands to 'query type' commands if zopen tools are read-only
prevent init, build, install, update-cacert, upgrade, remove, alt, clean from running if the person doesn't have write access to the root directory.

They 'can' run list, query, search.

MikeFultonDev:
Restrict alt, clean, init, install, remove, update-cacert to zopen tool chains that can be updated
fixes #463

### DIFF
--- a/bin/zopen-alt
+++ b/bin/zopen-alt
@@ -15,6 +15,7 @@ setupMyself()
         . "${INCDIR}/common.sh"
 }
 setupMyself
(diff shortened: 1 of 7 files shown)


IgorTodorovskiIBM (approved):
LGTM!

PR #466, opened 2023-09-27
Increase delay after github release is created

IgorTodorovskiIBM:
wait 30s so that it gets registered in github.com


MikeFultonDev (approved):
Kinda scary 😁
Is there a way to query ?

PR #467, opened 2023-09-27
Add checks to publish.groovy script

IgorTodorovskiIBM:
* Sometimes the github release is not reflected in github immediately after the release is created, so I added a loop that checks up to 5 times.
* Also, sometimes the release doesn't actually end up being created, so I make a last resort attempt and try to recreate it.
* This should also fix an issue with multiple .version files appearing, so it only look at installed .version now.

### DIFF
--- a/cicd/publish.groovy
+++ b/cicd/publish.groovy
@@ -1,3 +1,4 @@
+#!/bin/bash
 # Publish Job : https://163.74.88.212:8443/view/Framework/job/Port-Publish/
 # This publish job will a pax.Z artifact from the Port-Build (https://163.74.88.212:8443/view/Framework/job/Port-Build/)
 # Inputs: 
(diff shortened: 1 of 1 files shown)


cicd/publish.groovy
```diff
@@ -96,22 +95,64 @@ exists=$(github-release info -u ${GITHUB_ORGANIZATION} -r ${GITHUB_REPO}  -j)
...
+    exit 1                                     
+  fi
 fi
 
-sleep 30 # Let github register the release
-
-echo "Release should now exist"
-github-release info -u ${GITHUB_ORGANIZATION} -r ${GITHUB_REPO} --tag "${TAG}" -j
+# check up to 5 times for the release to appear on github                                                     
+for i in {1..5}; do
```
MikeFultonDev, line 129: not sure how expensive the call is but would it better to go with maybe 2 seconds and do it 25 times? 
if 10s is the 'sweet spot', then i am fine with this.

IgorTodorovskiIBM: From prior experience, it usually takes about 5-10s for the release to be reflected in github, so I'll stick with 10s for now and see how it goes.

PR #476, opened 2023-10-03
Fix zopen-split-patch - fails with file argument

IgorTodorovskiIBM:


MikeFultonDev:
oh - yeah - i manually did that yesterday - thanks!

PR #477, opened 2023-10-04
More fixes to replaceHardcodedPath logic, add sed as a dep

IgorTodorovskiIBM:
This fixes the issues found in texinfo (makeinfo). The z/OS /bin/sed has limitations in the regexp substitution, so I've added sed as a implicit dependency.


MikeFultonDev (approved):
LGTM

PR #482, opened 2023-10-05
Add github status check for shell check

IgorTodorovskiIBM:
- shell check: https://github.com/koalaman/shellcheck
- gives warnings and suggestions for bash/sh shell scripts
- Once we fix all of the issues, we can make it a mandatory status check for PRs

### DIFF
--- a/.github/workflows/shellcheck.yml
+++ b/.github/workflows/shellcheck.yml
@@ -1,0 +1,25 @@
+name: shellcheck
+
+on:
+  pull_request:
+    types: [opened, synchronize]
+    paths:
+      - '**.sh'
+      - '**zopen-*'
+      - '**/zopen'
+  push:
+    branches:
+      - main
+    paths:
+      - '**.sh'
+      - '**zopen-*'
+      - '**/zopen'
+
+jobs:
+  shellcheck-check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Lint check
+        uses: azohra/shell-linter@v0.6.0


DevonianTeuchter (approved):
LGTM.

MikeFultonDev (approved):
LGTM

IgorTodorovskiIBM:
Found this: https://github.com/marketplace/actions/sh-checker it includes shfmt as well as shellcheck

v1gnesh:
Wondering if shfmt could be used as a pre-commit hook instead.
This way, we get to see what changes it'll make before committing.
Also, VS Code and Jetbrains IDEs have integration to set this up.

https://github.com/scop/pre-commit-shfmt/tree/main

Similarly, both these IDEs have plugins for shellcheck; so can offer corrective actions while developing.
Correction is one click away. Action for shellcheck is still needed as it produces useful review comments.
As time goes and all zopen devs use shellcheck and shfmt locally (ideally), this Action can serve to make it easy for newer contributors and as a general fallback.

MikeFultonDev:
That makes sense. 
Are people comfortable it won’t introduce any bashisms? I don’t want to pre-req bash (yet at least) for our scripts.

IgorTodorovskiIBM:
A pre-commit hook for shfmt is a good idea

> That makes sense. Are people comfortable it won’t introduce any bashisms? I don’t want to pre-req bash (yet at least) for our scripts.

It's just a formatter so it should be ok, all the tools seem to work okay from my tests. `shellcheck` on the other hand can change the semantics of the program as I've unfortunately experienced, so I've had to disable some options here https://github.com/ZOSOpenTools/meta/pull/486/files#diff-3648cd4f7d46f434c6e6e21a44d23f5716574948ed690d79452535f1abf2ddb8R50

```
Printer options:

  -i,  --indent uint       0 for tabs (default), >0 for number of spaces
  -bn, --binary-next-line  binary ops like && and | may start a line
  -ci, --case-indent       switch cases will be indented
  -sr, --space-redirects   redirect operators will be followed by a space
  -kp, --keep-padding      keep column alignment paddings
  -fn, --func-next-line    function opening braces are placed on a separate line
```

v1gnesh:
For vim users, there are plugins for both shfmt and shellcheck.
https://github.com/z0mbix/vim-shfmt
https://github.com/itspriddle/vim-shellcheck
I haven't tested either..
Also, vim 8.x onwards comes with an in-built plugin manager, so no other dependencies needed.

MikeFultonDev (approved):
did not mean to approve...

MikeFultonDev (commented):
just a draft - ignore my approval 😁

PR #483, opened 2023-10-05
Fix bar chart and pie chart in progress page

IgorTodorovskiIBM:
Make not built grey in color and fixes the bar chart so that it is legible now


MikeFultonDev (approved):
Wonderful. LGTM

PR #494, opened 2023-10-12
Add in logic to only run cleanup

DevonianTeuchter:
commands when the main zopen
process.
Move the install mutex to reduce cleanup
operations without any real impact


MikeFultonDev (approved):
Lgtm

PR #496, opened 2023-10-13
Update README for 'using' info

ijmitch:


MikeFultonDev (approved):
good catch - thank you.

PR #497, opened 2023-10-13
Add logic to handle test harness

DevonianTeuchter:
Add logic to handle test harness failure (with ps not returning data?!?)
Move logic to ensure the pipeline terminates


MikeFultonDev (approved):
LGTM.

PR #502, opened 2023-10-13
500 urgent zopen clean u pkg too aggressive

DevonianTeuchter:
The test for isPackageActive does not correlate with the usage in zopen clean
Implement using new getter function for package version


MikeFultonDev (approved):
LGTM

PR #505, opened 2023-10-16
Fix OCI artifact generation

IgorTodorovskiIBM:


cicd/publish_oci.sh
```diff
@@ -52,8 +52,24 @@ LABEL vendor=\"ZOSOpenTools\""
...
+  exit 1
+fi
+
+set -e
+cd "\$INSTALL_DIR/${DIR_NAME}"
+./setup.sh
+ZZ
+
+  cat << ZZ > .zpm/.env
+cd "\$INSTALL_DIR/${DIR_NAME}";
```
MikeFultonDev, line 69: do you want to protect this better? What if the .env did a cd as well?

MikeFultonDev: never mind... i mis-read and thought it was . ./setup.sh

MikeFultonDev (approved):
LGTM

IgorTodorovskiIBM:
Added runtime dependencies as a label. Once the spec if official, we will modify accordingly

PR #509, opened 2023-10-17
clean up options and put into sections

MikeFultonDev:
please review - in particular the stuff I pushed down into 'deprecated' ...

### DIFF
--- a/bin/zopen-build
+++ b/bin/zopen-build
@@ -55,48 +55,30 @@ printEnvVar()
 {
   echo "
 User-Provided environment variables:
-  ZOPEN_CC             C compiler (default set by dependency)
-  ZOPEN_CXX            C++ compiler (default set by dependency)
-  ZOPEN_CPPFLAGS       C/C++ pre-processor flags (default set by dependency)
-  ZOPEN_CFLAGS         C compiler flags (default set by dependency)
-  ZOPEN_CXXFLAGS       C++ compiler flags (default set by dependency)
-  ZOPEN_LDFLAGS        C/C++ linker flags (default set by dependency)
-  ZOPEN_LIBS           C/C++ libraries (default set by dependency)
-  ZOPEN_TYPE           (Deprecated) The type of package to download. Valid types are TARBALL, BARE and GIT (required)
-  ZOPEN_TARBALL_URL    (Deprecated) The fully qualified URL that the tarball should be downloaded from (required if ZOPEN_TYPE=TARBALL)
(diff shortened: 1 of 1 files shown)


DevonianTeuchter (approved):
LGatM

bin/zopen-build
```diff
@@ -55,72 +55,96 @@ printEnvVar()
...
+  ZOPEN_TYPE           The type of package to download. Valid types are TARBALL, BARE and GIT 
+  ZOPEN_TARBALL_URL    The fully qualified URL that the tarball should be downloaded from (required if ZOPEN_TYPE=TARBALL)
+  ZOPEN_TARBALL_DEPS   Space-delimited set of source packages this git package depends on to build (required if ZOPEN_TYPE=TARBALL)
+  ZOPEN_GIT_URL        The fully qualified URL that the git repo should be cloned from (required if ZOPEN_TYPE=GIT)
+  ZOPEN_GIT_DEPS       Space-delimited set of source packages this tarball package depends on to build (required if ZOPEN_TYPE=GIT)
+  ZOPEN_GIT_BRANCH     The branch that the git repo should checkout 
+  ZOPEN_GIT_TAG        The tag that the git repo should checkout as a branch 
+  ZOPEN_URL            Alternate environment variable instead of ZOPEN_TARBALL_URL or ZOPEN_GIT_URL (alternate to ZOPEN_TARBALL_URL or ZOPEN_GIT_URL)
+  ZOPEN_DEPS           Alternate environment variable instead of ZOPEN_TARBALL_DEPS or ZOPEN_GIT_DEPS (alternate to ZOPEN_TARBALL_DEPS or ZOPEN_GIT_DEPS)
+  ZOPEN_DEV_TYPE       The type of package to download. Valid types are TARBALL, BARE and GIT
```
IgorTodorovskiIBM, line 131: This is optional and in case the "heuristic" does not detect the type correctly, this can force it to the right type.

MikeFultonDev: 'this' being ZOPEN_xxx_TYPE right? I moved this to deprecated because none of the repos actually use it.

MikeFultonDev: If you think it's useful but currently unused, then I can move those 2 up to that.

IgorTodorovskiIBM: Actually I think we'll need it for https://github.com/ZOSOpenTools/comp_clangport/blob/main/buildenv and the other compiler stubs once we migrate to the STABLE/DEV buildenv format or perhaps we can add logic to set it to BARE (or equivalent) if no url is specified

MikeFultonDev: moved to a new section indicating currently unused.

IgorTodorovskiIBM (approved):
LGTM

PR #510, opened 2023-10-17
Updates for quickstart

netskink:
# Addition for upgrade

Uses Igor's upgrade method

# Addition for trouble shooting in case of pinned

Uses Mike's method for fixing the problem.

### DIFF
--- a/docs/Guides/QuickStart.md
+++ b/docs/Guides/QuickStart.md
@@ -28,3 +28,73 @@ If you like learning through watching, check out our getting started Video:
 <video height="1080" controls>
   <source src="https://github.com/ZOSOpenTools/collateral/raw/main/ZOSOpenToolsIntroV2-cropped.mp4" type="video/mp4">
 </video>
+
+
+## Upgrading z/OS Open Tools
+
+z/OS Open Tools can be upgraded like so:
+
+```
+$ zopen upgrade
+```
+
+In this case, all packages including the package manager, `zopen` (via the meta package manager) will
+be updated sequentially while being prompted to perform the update.
+Using the option `-y` will automatically accept all prompts as
+shown here:
+
+```
+$ zopen upgrade -y
+```
+
+
+## Upgrading a single package
+
+Individual packages can be upgraded by specifying the package
+name.  For example upgrading just the package manager is
+accomplished like so:
+
+```
+$ zopen upgrade
+```
+
+In this case, all packages including the package manager (Meta) will
+be updated sequentially while being prompted to perform the update.
+Using the option `-y` will automatically accept all prompts as
+shown here:
+
+```
+$ zopen upgrade -y
+```
+
+## Upgrading a single package
+
+Individual packages can be upgraded by specifiying the package
+name.  For example upgrading just the package manager is
+accomplished like so:
+
+```
+$ zopen upgrade meta -y
+$ zopen init --re-init
+# This assumes zopen is installed in your $HOME directory
+$ . ~/zopen/etc/zopen-config
+```
+
+Note, for users of meta 0.6.x and earlier, if the upgrade fails because of a pinned message, find the 
+`.pinned` file and remove it.  Afterwards `zopen upgrade meta -y` will
+work.
+
+For example:
+
+```
+$ cd ~/zopen
+$ find . -name ".pinned"
+./meta-0.6.2/.pinned
+./usr/local/zopen/meta/meta-dt/.pinned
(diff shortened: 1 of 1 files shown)


DevonianTeuchter (commented):
Typos on lines 38 & 65.
Line 42 should state the instructions are for explicitly upgrading the package manager - meta is treated as any other package so actually gets updated by the command above (line 35) also - the breakdown as-is suggests the package manager has to be updated separately.

netskink:
> Typos on lines 38 & 65. Line 42 should state the instructions are for explicitly upgrading the package manager - meta is treated as any other package so actually gets updated by the command above (line 35) also - the breakdown as-is suggests the package manager has to be updated separately.

Yes, indeed.  Many thanks!

Regarding the separation.  I thought it had to be separate.  Is that not the case?  When I do `zopen upgrade` it did everything but meta.  Later, Mike told me about the second command.  At that time, it failed with meta is pinned, which I suppose was the result of installing zopen for the first time.  (I had saved off my original zopen dir to archive).

Once I removed the .pinned file it was able to update meta.

docs/Guides/QuickStart.md
```diff
@@ -28,3 +28,46 @@ If you like learning through watching, check out our getting started Video:
...
+```
+$ zopen upgrade
+```
+
+
+## Upgrading the zopen package manager
+
+Use this process to upgrade the zopen package manager (meta).
+By default, `zopen upgrade` updates the individual packages.
+In order to update the package manager (meta), it needs
```
MikeFultonDev, line 46: i don't believe the package manager has to be done separately so i think you can safely delete this sentence. @DevonianTeuchter please comment if otherwise

DevonianTeuchter: Yes, can be done at the same time - only reason for it being called out was when upgrading from a pinned (pre-fork-merge) version where .pinned was set for meta.

netskink: Ok. I will try to capture the situation.  Please review the next update.

docs/Guides/QuickStart.md
```diff
@@ -28,3 +28,54 @@ If you like learning through watching, check out our getting started Video:
 <video height="1080" controls>
   <source src="https://github.com/ZOSOpenTools/collateral/raw/main/ZOSOpenToolsIntroV2-cropped.mp4" type="video/mp4">
 </video>
+
+
+## Upgrading zopen 
+
+Zopen can be upgraded like so:
```
IgorTodorovskiIBM, line 35:
```suggestion
z/OS Open Tools can be upgraded like so:
```
To indicate that the tools can be updated vs the zopen tool itself

docs/Guides/QuickStart.md
```diff
@@ -28,3 +28,54 @@ If you like learning through watching, check out our getting started Video:
...
+
+```
+$ zopen upgrade
+```
+
+In this case, all packages including the package manager (Meta) will
```
IgorTodorovskiIBM, line 41:
```suggestion
In this case, all packages including the package manager, zopen (via the meta package) will
```

docs/Guides/QuickStart.md
```diff
@@ -28,3 +28,54 @@ If you like learning through watching, check out our getting started Video:
...
+shown here:
+
+```
+$ zopen upgrade -y
+```
+
+
+## Upgrading a single package
+
+Individual packages can be upgraded by specifiying the package
```
IgorTodorovskiIBM, line 53:
```suggestion
Individual packages can be upgraded by specifying the package
```

docs/Guides/QuickStart.md
```diff
@@ -28,3 +28,54 @@ If you like learning through watching, check out our getting started Video:
...
+accomplished like so:
+
+```
+$ zopen upgrade meta -y
+$ zopen init --re-init
+# This assumes zopen is installed in your $HOME directory
+$ . ~/zopen/etc/zopen-config
+```
+
+Note, if the upgrade fails because of a pinned message, find the
```
IgorTodorovskiIBM, line 64:
```suggestion
Note, for users of meta 0.6.x and earlier, if the upgrade fails because of a pinned message, find the 
```

IgorTodorovskiIBM (approved):
LGTM

HarithaIBM (approved):

netskink:
Thanks for approving it.  Please don't merge yet.  I'm going to try to squash some commits prior to merge on your part.


I view this part as more substantial since I can use this as a method to make cleaner commits to a moving upstream.

PR #512, opened 2023-10-18
force zopen_get_version to be required

MikeFultonDev:


bin/zopen-build
```diff
@@ -1502,7 +1502,7 @@ install()
     fi
 
     if [ -z "$(command -v zopen_get_version)" ]; then
-      printWarning "A zopen_get_version() function should be defined in the buildenv to extract the version contents"
+      printError "A zopen_get_version() function must be defined in the buildenv to extract the version contents"
```
IgorTodorovskiIBM, line 1505: Minor: I wonder if we should catch this early rather than going through an entire build to witness it?

MikeFultonDev: https://github.com/ZOSOpenTools/meta/issues/516 opened to tackle this (we should do all functions this way - e.g. also the check function)

PR #514, opened 2023-10-18
Check if JSON cache is accessible

DevonianTeuchter:
Gracefully fail if the cache files for the JSONrequest logic are not read or writeable.


IgorTodorovskiIBM (approved):
LGTM

PR #517, opened 2023-10-18
Fix zoslib env hook so that it only avoids implicitly setting envars for tool it is building

IgorTodorovskiIBM:
The issue is that the zoslib env hook logic wasn't active when we're doing a zopen build. We added this logic because at check/test time, the tests were failing because the environment variables were pointing to bad paths.

But this also meant that any dependent tools that set envars via zoslib env hooks would avoid setting those environment variables, effectively breaking them.

----

Also adds overrideset action to force set environment variables which should not be customized

### DIFF
--- a/bin/zopen-build
+++ b/bin/zopen-build
@@ -1807,7 +1807,11 @@ char* value_str;
 char* pos;
 long size;
 
-// Do not use when building itself to avoid test failures to incorrect paths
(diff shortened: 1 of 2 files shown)


MikeFultonDev (approved):
looks good - do you want to add a couple more lines in the comment? this is pretty subtle

MikeFultonDev (approved):

PR #519, opened 2023-10-19
remove partially implemented verbs - add back in when tested

MikeFultonDev:


IgorTodorovskiIBM:
Alternative would be to list a "beta" category of commands that we would like to get feedback on

IgorTodorovskiIBM (approved):

MikeFultonDev:
> Alternative would be to list a "beta" category of commands that we would like to get feedback on

we have open issues for basic options processing so once those are fixed, i'd be ok with that.

PR #520, opened 2023-10-19
help2man wants executable name, not full-path

MikeFultonDev:


MikeFultonDev:
with this fix, help2man no longer hangs

PR #524, opened 2023-10-19
Add zoslib as an implicit dep

IgorTodorovskiIBM:
If we encounter builds that don't work well with zoslib, we can set ZOPEN_DONT_ADD_ZOSLIB_DEP.


bin/zopen-build
```diff
@@ -87,6 +87,7 @@ Optional:
   ZOPEN_INSTALL_OPTS   Options to pass to installation program (defaults to '${ZOPEN_INSTALL_OPTSD}')
   ZOPEN_CLEAN          Clean up program to run (defaults to '${ZOPEN_CLEAND}')
   ZOPEN_CLEAN_OPTS     Options to pass to clean up  program (defaults to '${ZOPEN_CLEAN_OPTSD}')
+  ZOPEN_DONT_ADD_ZOSLIB_DEP Set it to avoid adding zoslib as a dependency
```
MikeFultonDev, line 90: can you put it create a line above it about it's restricted usage so people will understand they probably don't want to use it?

MikeFultonDev (approved):

PR #530, opened 2023-10-20
set non-zero rc when help2man fails to generate file

MikeFultonDev:
make sure check fails if any man pages can't be generated

### DIFF
--- a/bin/zopen-help2man
+++ b/bin/zopen-help2man
@@ -62,6 +62,7 @@ fi
 
 tooldirs="${rootdir}/bin"
 
+rc=0
 echo "Man pages will be written to: ${mandir}"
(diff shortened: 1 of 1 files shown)


IgorTodorovskiIBM (approved):
LGTM

DevonianTeuchter (commented):
Message on line 76 is missing the text  "--version." .  Since it needs editing anyway, would matching the format of the message on line 70 be consistent...?

HarithaIBM (approved):

MikeFultonDev:
> Message on line 76 is missing the text "--version." . Since it needs editing anyway, would matching the format of the message on line 70 be consistent...?

The messages are different because there are different problems - it could be missing one of --version, --help, or both.

PR #532, opened 2023-10-20
Do not add curl path to PATH when updating dependencies

IgorTodorovskiIBM:
With the new zopen, the curl path is the path containing all z/OS Open Tools.
As such, all z/OS Open Tools that are currently installed on the system are used which leads to unpredictable results, depending on what is installed. We should only include the deps specified in the `ZOPEN_*_DEPS` list.


MikeFultonDev (approved):
Lgtm

PR #536, opened 2023-10-21
Add references in zopen to other commands 

MikeFultonDev:
For regular man users, this says what other docs there are and for info users, they can click to the other man page


IgorTodorovskiIBM (approved):
Lgtm

PR #537, opened 2023-10-22
Fix issue with zoslib_env_hook prepend action when PROJECT_ROOT specified more than once

IgorTodorovskiIBM:
After changing PER5LIB from set to prepend, this was leading to segmentation faults. This fixes the issue


MikeFultonDev (approved):
Subtle bug. Lgtm

PR #539, opened 2023-10-24
Improved Explicit Upgrade Message

DangerousDB:
Closes #429 This solves issue #429 by adding explanation as to why a package with a newer release is having its upgrade skipped, and why a package is not being reinstalled when it is the requested version


DevonianTeuchter (changes requested):
The change to the "skip" message doesn't necessarily correlate with the user action.  
If they specify "zopen upgrade git", then ncurses will be checked (as a dependency of git) and will display the "...due to the use of the skipupgrade flag..." -  but the user never specified that flag so will have no frame of reference as to why that is mentioned in the message.
Perhaps something more succinct like
"Package <ver> has a newer version <ver> available but was not specified for upgrade"...?

DangerousDB:
Ok, I have fixed it

DevonianTeuchter (approved):

PR #542, opened 2023-10-27
Rework cleanup logic

DevonianTeuchter:
Regenerate trap commands everytime a new command is added rather than
maintain a list (in file or pipe)
"Simplify" progress handler (remove duplicate code) and add logic to detect if orphaned

### DIFF
--- a/include/common.sh
+++ b/include/common.sh
@@ -3,42 +3,60 @@
 # shellcheck shell=sh
 #
 
-addCleanupTrapCmd()
+zopenInitialize()
 {
-  newCmd=$(echo "$1" | sed -e 's/^[ ]*//' -e 's/[ ]*$//')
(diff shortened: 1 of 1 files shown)


include/common.sh
```diff
@@ -790,7 +792,7 @@ printError()
   printColors "${NC}${RED}${BOLD}***ERROR: ${NC}${RED}${1}${NC}" >&2
   [ -n "${xtrc}" ] && set -x
   mutexFree "zopen" # prevent lock from lingering around after an error
-  cleanupFunction
+cleanupFunction
```
IgorTodorovskiIBM, line 795: Is this change intended?

DevonianTeuchter: No, just me not spotting the indentation change as "show whitespace differences" wasn't enabled when I reinstated the cleanup call (as was testing without it)

Issue #202, opened 2023-02-22
bold yellow impossible to read on white background


MikeFultonDev:
Request to either switch the colour or detect the background and intelligently swap

![image](https://user-images.githubusercontent.com/61100689/220517580-c70bcbc8-c773-4adb-aee2-c3a8b226efc1.png)

MikeFultonDev:
What if we went with a 'colour free' approach (probably more likely to work for everyone), i.e.

**bold underline** <-- error
**bold** <-- warning

??

IgorTodorovskiIBM:
What about headers? bold too?

Here's the output from brew:
<img width="1363" alt="Screenshot 2023-02-22 at 9 15 58 PM" src="https://user-images.githubusercontent.com/39890068/220807281-f6d62d25-4368-45fe-8b16-1424d1230320.png">

IgorTodorovskiIBM:
For warnings:
<img width="806" alt="Screenshot 2023-02-22 at 9 18 28 PM" src="https://user-images.githubusercontent.com/39890068/220807405-0d6a55d8-fa6d-4d45-9049-a8eabadcfd86.png">

IgorTodorovskiIBM:
<img width="884" alt="Screenshot 2023-02-22 at 9 19 31 PM" src="https://user-images.githubusercontent.com/39890068/220807510-75e1653e-ebb2-4bd2-8a71-f1d39c42beb8.png">
They don't handle it differently when the background is white, but the yellow they've chosen is a darker yellow

IgorTodorovskiIBM:
<img width="581" alt="Screenshot 2023-02-22 at 9 31 15 PM" src="https://user-images.githubusercontent.com/39890068/220808890-cc6ab768-960a-4e66-8282-1d0c583a4c75.png">
Red for error

IgorTodorovskiIBM:
I wouldn't mind following the same approach as brew:
* Bold for headers
* Red for errors
* Yellow for warnings

MikeFultonDev:
Works for me. What does clang do?

MikeFultonDev:
We can also check the background (figured out how to do this) so that we can be like emacs and choose reasonable colors based on the background (at least when TERM=xterm):

```
# 
# return 1 if brightness is 'dark', 0 if brightness is 'light', -1 if darkness unknown
# 
function darkbackground() {
  if [ "${TERM}x" = "xtermx" ]; then
    printf '\033]11;?\007'; read -N 24 bg
[...]

PR #543, opened 2023-10-27
Update common.sh to allow for a "light mode"

Thalia-J:
This allows people with white backgrounds to read warnings and headers by exporting their COLORFGBG in bash to change them to magenta instead of yellow. 

example: 
export COLORFGBG="0;15" for light backgrounds
export COLORFGBG="0;0" for dark backgrounds

I left the colors readable in both dark and light mode unchanged.

This closes issue #202

### DIFF
--- a/include/common.sh
+++ b/include/common.sh
@@ -232,7 +232,10 @@ defineANSI()
   ERASELINE="${ESC}[2K"
   CRSRHIDE="${ESC}[?25l"
   CRSRSHOW="${ESC}[?25h"
-
+  
+  
+  
+  
   # Color-type codes, needs explicit terminal settings
   if [ ! "${_BPX_TERMPATH-x}" = "OMVS" ] && [ -z "${NO_COLOR}" ] && [ ! "${FORCE_COLOR-x}" = "0" ] && [ -t 1 ] && [ -t 2 ]; then
     esc="\047"
@@ -262,6 +265,18 @@ defineANSI()
     BOLD=''
     UNDERLINE=''
     NC=''
+  fi
+
+  darkbackground
+  bg=$?
+  if [ $bg -ne 0 ]; then 
+    #if the background was set to black or unknown the header and warning color will be yellow
+    HEADERCOLOR="${YELLOW}"
+    WARNINGCOLOR="${YELLOW}"
+  else
+    #else the header and warning color will become magenta
+    HEADERCOLOR="${MAGENTA}"
+    WARNINGCOLOR="${MAGENTA}"
   fi
 }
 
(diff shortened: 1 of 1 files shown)


DevonianTeuchter (changes requested):
The gist of the changes seem fine. However by moving the [ -z "${-....  checks inside the if..then block, when -x/xdebug trace is enabled these functions will become noisy again. It'd be better to have the [ -z ... tests as the first call in the function leaving [ ! -z... prior to the returns.
Also the  [ $dark -eq -1 ] || [ $dark -eq 1]    check might be better as
[ $dark -ne 0 ] since the possible options are -1:0:1 .

The other issue is that this adds a lot of overhead/complexity to these methods.  With a wee bit of refactoring, the complexity can be contained:
defineAnsi() could contain the new "dark" code and setting ${HEADER} and ${WARNING} to either YELLOW or MAGENTA  (possibly even ${ERROR}->${RED})
- the printHeader function could then call printColors "${NC}${HEADER}${BOLD}${UNDERLINE{${1}${NC}" >&2
- printWarning similarly calling printColors "${NC}${WARNING}${BOL...

include/common.sh
```diff
@@ -649,8 +663,17 @@ printVerbose()
 
 printHeader()
 {
+  darkbackground
+  dark=$?
+  if [ $dark -eq -1 ] || [ $dark -eq 1]; then
```
IgorTodorovskiIBM, line 668:
```suggestion
  if [ $dark -eq -1 ] || [ $dark -eq 1 ]; then
```

Thalia-J:
> The gist of the changes seem fine. However by moving the [ -z "${-.... checks inside the if..then block, when -x/xdebug trace is enabled these functions will become noisy again. It'd be better to have the [ -z ... tests as the first call in the function leaving [ ! -z... prior to the returns. Also the [ $dark -eq -1 ] || [ $dark -eq 1] check might be better as [ $dark -ne 0 ] since the possible options are -1:0:1 .
> 
> The other issue is that this adds a lot of overhead/complexity to these methods. With a wee bit of refactoring, the complexity can be contained: defineAnsi() could contain the new "dark" code and setting ${HEADER} and ${WARNING} to either YELLOW or MAGENTA (possibly even ${ERROR}->${RED})
> 
> * the printHeader function could then call printColors "${NC}${HEADER}${BOLD}${UNDERLINE{${1}${NC}" >&2
> * printWarning similarly calling printColors "${NC}${WARNING}${BOL...

From this new commit: 8f52464499975074d89e4276b106faa15efe4706
I took your advice and made the colors be updated in the defineANSI() function. It updates HEADERCOLOR and WARNINGCOLOR directly through those variables.

MikeFultonDev:
@DevonianTeuchter and @IgorTodorovskiIBM and @Thalia-J is this good to merge now?

include/common.sh
```diff
@@ -211,14 +211,31 @@ deref()
...
+
 defineANSI()
 {
   # Standard tty codes
   ESC="\047"
   ERASELINE="${ESC}[2K"
   CRSRHIDE="${ESC}[?25l"
   CRSRSHOW="${ESC}[?25h"
-
+
```
IgorTodorovskiIBM, line 235: I'd recommend deleting these newlines

IgorTodorovskiIBM (approved):
LGTM, minor comment

include/common.sh
Thalia-J: 659e5746f44520595d5c06bf011200e2409144f0
Removed the newlines

DevonianTeuchter (approved):
LGTM

PR #548, opened 2023-10-31
Tag and encode .env as IBM-1047

IgorTodorovskiIBM:
The .env is the entry point for any consumer/developer and as such, should be encoded in the default encoding (IBM-1047). This way we can also add the required auto-conversion environment variables

### DIFF
--- a/.env
+++ b/.env
@@ -1,6 +1,9 @@
-# Setup no longer required if you are just 'using' tools 
-# It may be convenient if you are 'developing' tools
+# Must be sourced as follows: . ./.env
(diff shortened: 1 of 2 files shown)


MikeFultonDev (commented):
Looks good. Since this is a very special marking of IBM-1047 for people sourc'ing from
a dubious shell, do you want to add a comment to that effect to indicate this is why you mark
it 1047?

MikeFultonDev (approved):
Lgtm

IgorTodorovskiIBM:
> Looks good. Since this is a very special marking of IBM-1047 for people sourc'ing from a dubious shell, do you want to add a comment to that effect to indicate this is why you mark it 1047?

Added a comment, thanks

MikeFultonDev (approved):
LGTM

PR #551, opened 2023-11-02
Update ansible to reflect new zopen

IgorTodorovskiIBM:
* Changed region from tokyo to toronto (doesn't experience the curl timeout issues we saw with Tokyo based instances)
* Updates to the latest wazi images
* Adds fixes for using python 11 which is now available on the wazi images (adds /lib to LIBPATH for libzz64)
* Modifies the code to extra meta pax and install zopen, including git and bash (implicit from git)
* Copies a working private key so that git clone on the ssh protocol work on the system


MikeFultonDev (approved):
LGTM

PR #559, opened 2023-11-08
Clean up zopen-xxx services and associated docs

MikeFultonDev:
zopen-versions fixed so it now works
zopen-query cleaned up so the documented functions work - others highlighted for future improvements
zopen doc page fixed to point to the right underlying tools and some wordsmithing


IgorTodorovskiIBM (approved):
LGTM

Issue #468, opened 2023-09-27
Add support when there is no tty available


IgorTodorovskiIBM:
We use stty to get the columns value, but stty fails because there is not tty available. We shouldn't need a tty when requesting an install

From @ccw-1 
```
- Querying for installed packages
stty: !FSUMB045 failed to get termios attributes: EDC5137I Inappropriate I/O control operation.
printInstalledEntries 3: /home/ccw/meta/bin/zopen-query 380: FSUM6659 syntax error in arithmetic expression " / 4 -1 " near /
- Querying remote repo for latest package information
- ```

**Comment by MikeFultonDev** [comment]
When does no tty exist?

**Comment by IgorTodorovskiIBM** [comment]
> When does no tty exist? Is this via 'ssh' typically @IgorTodorovskiIBM ?

Yes, this was emite via the `ssh -T` option.

Here are the steps to reproduce:
```
itodorov@rogi21:~$ ssh -T itodoro@zoscan2b.pok.stglabs.ibm.com "export _BPXK_AUTOCVT=ON; . zopen/etc/zopen-config; zopen list --installed"
Processing z/OS Open Tools configuration...DONE
stty: !FSUMB045 failed to get termios attributes: EDC5137I Inappropriate I/O control operation.
printInstalledEntries 4: /home/itodoro/zopen/usr/local/bin/zopen-query 385: FSUM6659 syntax error in arithmetic expression " / 4 - 1" near /
```

IgorTodorovskiIBM:
I can look into it

PR #560, opened 2023-11-08
Check if tty is available when getting columns + add other ways to obtain columns

IgorTodorovskiIBM:
Fixes https://github.com/ZOSOpenTools/meta/issues/468

### DIFF
--- a/include/common.sh
+++ b/include/common.sh
@@ -319,7 +319,7 @@ getScreenCols()
   if  [ -t 1 ] && [ -t 2 ]; then
     # Note tput does not handle ssh sessions too well...
     stty | awk -F'[/=;]' '/columns/ { print $4}' | tr -d " "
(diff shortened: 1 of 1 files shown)


include/common.sh
```diff
@@ -316,8 +316,14 @@ ansiline()
 
 getScreenCols()
 {
-  # Note tput does not handle ssh sessions too well...
-  stty | awk -F'[/=;]' '/columns/ { print $4}' | tr -d " "
+  if  [ -t 1 ] && [ -t 2 ]; then
```
MikeFultonDev, line 319: I had to look this up - might just be me - but would it make sense to have a comment that says that -t fd tests if the descriptor is opened on a terminal?

IgorTodorovskiIBM: thanks, added a comment

MikeFultonDev (approved):
LGTM

PR #564, opened 2023-11-08
Provide the total size up front when installing all tools (zopen install --all)

IgorTodorovskiIBM:
Output as of today:
```
You have chosen to install all tools. An estimated 2819.17 MB of additional disk space will be used.
```

### DIFF
--- a/bin/zopen-install
+++ b/bin/zopen-install
@@ -623,9 +623,14 @@ installArray=""
 if ${all}; then
   if ! ${yesToPrompts}; then
     # Sum up the pax size + expanded size of the latest releases of each port
-    allSize=$(jq '.release_data | to_entries | map(select(.value | length > 0)) | map(.value[0].assets[0]) | reduce .[] as $item ({}; . + {($item.name): ($item.size + $item.expanded_size)}) | [.[]] | add' ${JSON_CACHE})
-    megabytes=$(echo "scale=2; ${allSize} / (1024 * 1024)" | bc)
-    printInfo "You have chosen to install all tools. An estimated ${megabytes} MB of additional disk space will be used."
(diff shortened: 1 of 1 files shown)


DevonianTeuchter (commented):
LGTM as-is but would a 'du' space check be a useful addition at this point to save failure during the install if insufficient space?

MikeFultonDev (approved):
LGTM - if you haven't already, do you want to open an issue to handle the size calculation for a set of packages (as opposed to all)?

MikeFultonDev:
i like that. i also think if there is less than XX MB in the file system, we should just fail too... If it's close enough to empty, they should clean up.

IgorTodorovskiIBM:
I like the idea as well! However, I think some filesystems are set up such that they can grow once they reach the max quota. Should we always emit an error? For now I've added it as a warning.

```
You have chosen to install all tools. An estimated 2840 MB of additional disk space will be used.
***WARNING: Your zopen file-system (/home/itodoro/zopen) only has 260 MB of available space.
Enter 'all' to confirm full installation [takes a long time so be sure!]:
```

DevonianTeuchter:
> I like the idea as well! However, I think some filesystems are set up such that they can grow once they reach the max quota. Should we always emit an error? For now I've added it as a warning.
> 
> ```
> You have chosen to install all tools. An estimated 2840 MB of additional disk space will be used.
> ***WARNING: Your zopen file-system (/home/itodoro/zopen) only has 260 MB of available space.
> Enter 'all' to confirm full installation [takes a long time so be sure!]:
> ```
LGTM - A warning seems reasonable - we've advised them, they can answer no and if they still proceed then that's on them - and like you say, some filesystems can dynamically resize.

DevonianTeuchter (approved):

PR #570, opened 2023-11-09
Fix metadata.json creation for BARE type projects

IgorTodorovskiIBM:
We need the base (implicit) dependencies, in particular jq, for metadata.json creation.

This will fix the builds for comp_clang/go/xlclang.


MikeFultonDev (approved):
LGTM - good catch

PR #574, opened 2023-11-10
Use active symlink for LICENSE

DevonianTeuchter:
While creating new tests, zopen_generate would fail to find the meta LICENSE file.
The lookup for that file is to: ${ZOPEN_ROOTFS}/usr/local/zopen/meta/meta-main/LICENSE  however I had a different location for my meta as I was testing it.  There should (always?) be a link from meta/meta to the active versioned release which is constant, rather than meta-main which might not exist.  This PR proposes using the main symlink to find metas LICENSE

### DIFF
--- a/bin/zopen-generate
+++ b/bin/zopen-generate
@@ -238,7 +238,7 @@ install/
 EOT
 printInfo "${name}port/.gitignore created"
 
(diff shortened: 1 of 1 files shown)


IgorTodorovskiIBM (approved):
Good catch, LGTM

PR #577, opened 2023-11-14
Updating Docs to reference white background color command

Thalia-J:
Added in documentation for the color scheme change command in merged commit 8263fc1

### DIFF
--- a/docs/Guides/CommonSolutions.md
+++ b/docs/Guides/CommonSolutions.md
@@ -216,3 +216,10 @@ Internally, zopen sets the `V` and `VERBOSE` environment variables to 1 when the
 ```
 zopen build -vv
 ```
+
+# Some of the text colors are too bright for my terminal's background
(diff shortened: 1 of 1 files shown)


docs/Guides/CommonSolutions.md
```diff
@@ -216,3 +216,10 @@ Internally, zopen sets the `V` and `VERBOSE` environment variables to 1 when the
 ```
 zopen build -vv
 ```
+
+# Some of the text colors are too bright for my terminal's background
+If your background is a lighter color, the header or warning colors could be almost impossible to read as they are by default yellow. There exists a variable in bash that if you change it, it will update the color scheme. The following **bash** command is for changing the color scheme to fit a light terminal background.
```
IgorTodorovskiIBM, line 221: Also worth mentioning that many terminal emulators set this environment variable automatically.

HarithaIBM (approved):

docs/Guides/CommonSolutions.md
MikeFultonDev: @Thalia-J can you add a comment per @IgorTodorovskiIBM 's suggestion?

PR #581, opened 2023-11-15
Fix could not locate curl directory error if you have multiple curl installations

IgorTodorovskiIBM:
If we upgrade to newer versions of curl/jq, then there could be multiple boot installations

Fixes this issue:
```
***ERROR: Could not locate curl directory '/home/itodoro/zopen/boot/curl-8.1.2
curl-8.4.0/' for bootstrap. Re-run 'zopen init' to retry
```

Same fix applied to jq.


DevonianTeuchter (approved):
Lgtm

PR #587, opened 2023-11-20
zopen-build: Add an option to avoid installing runtime deps

IgorTodorovskiIBM:
By default runtime dependencies are installed as part of the zopen build process. Sometimes we don't want this, especially when we are debugging runtime dependencies along with the current project.


DevonianTeuchter (approved):
LGTM.

PR #588, opened 2023-11-21
add comments on code and fix compute of end of file line number

MikeFultonDev:
Once checked in, I will also have a new test for split-patches under metaport that I have already verified.
I can't check in the metaport fix until this goes down because it will fail without these fixes.


IgorTodorovskiIBM (approved):
LGTM

PR #589, opened 2023-11-21
Add gnport to list of packages for depot_toolsport

netskink:
I'm certain this is wrong, but I need it for depot tools.  Perhaps once zoslib dependency is resolved, gnport will build, and a binary with a tag will be present so this is done automatically.  If not, this is the start of the manual process. No rush on this pull request.


netskink:
ahh, this is done automatically. No need for this.

PR #590, opened 2023-11-21
Query bug

DevonianTeuchter:
Fix bug with querying for installed packages from within zopen install/upgrade.  zopen-install calls zopen-query directly rather than using the list verb for zopen; the parameters are different for when a "list" action is required - zopen uses the "list" verb, zopen-query uses the "--list" parameter.  zopen-install incorrectly specifies "list" rather than "--list" and as such, the list of returned packages can become corrupted.  For small setups, this might not be noticeable as the returned list is sanitised, however with a sufficiently long list (like if a "--all" had been used previously), the list of packages will become corrupt and break the upgrade mechanism as well as taking a longer period of time than it should.  Full installations might not be upgradeable using the "zopen upgrade" command, packages might need to be manually listed.
The find used to determine the installed packages also recurses too deeply, returning huge volumes of data - again, this gets sanitised for processing but takes a long period of time and can cause issues with envvar lengths etc.  The zosfind should only now pick up in the current directory after switching to PKGINSTALL
In addition, colorization of columns in detailed mode and column headers have been updated and various shellcheck issues resolved.

### DIFF
--- a/bin/zopen-query
+++ b/bin/zopen-query
@@ -61,7 +61,7 @@ colorizepct()
   elif [ ${percentage} -eq 100 ]; then
     colored="${GREEN}"
   elif [ ${percentage} -gt 50 ]; then
-    colored="${BLUE}"
+    colored="${BOLD}${BLUE}"
   elif [ ${percentage} -gt 0 ]; then
     colored="${YELLOW}"
   else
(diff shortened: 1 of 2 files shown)


bin/zopen-query
```diff
@@ -61,7 +61,7 @@ colorizepct()
   elif [ ${percentage} -eq 100 ]; then
     colored="${GREEN}"
   elif [ ${percentage} -gt 50 ]; then
-    colored="${BLUE}"
+    colored="${BOLD}${BLUE}"
```
IgorTodorovskiIBM, line 64: Is the blue more recognizable when bold?

IgorTodorovskiIBM (approved):
LGTM

bin/zopen-query
DevonianTeuchter: To my eyes, yes - it's brighter

PR #594, opened 2023-11-22
Message cleanup

AnthonyGiorgio:
A significant number of output messages did not have periods at the end of the text.  I've added them, except in the case where the last part of the message is a path.

I've cleaned up the verbose message output by removing extraneous single quotes around the text that cluttered things up.

I also fixed a variable (`chosenRepo`) that had the wrong case, which prevented the text from being printed.


IgorTodorovskiIBM (approved):
Thanks! LGTM

DevonianTeuchter (approved):
LGTM

PR #602, opened 2023-11-28
Add zopen_pre_terminate hook

IgorTodorovskiIBM:
It will run before zopen-build terminates. It also runs after the package contents are created so it has no impact on the install dir package


MikeFultonDev (approved):
Lgtm

PR #607, opened 2023-12-04
Analytics - Collecting usage statistics for zopen

IgorTodorovskiIBM:
* This PR adds code to collect usage statistics for zopen.

This feature will be enabled as a beta feature. Set ZOPEN_BETA_FEATURES=1 so that we can perform internal testing before we make it live to the general public.

* Starting point is zopen-init
  *  zopen-init adds logic to check if the user is running on an IBM host and enables usage statistics by default. If not an IBM host, then it prompts the user if they want to enable usage statistics.
  * You can also enable usage statistics using the option --enable-stats or to disable, --noenable-stats
  * For CI/CD bots, the environment variable ZOPEN_IS_BOT will be used (which we can set in our groovy scripts)
  * For build install, the environment variable ZOPEN_IN_ZOPEN_BUILD is used to determine if the install is from a build
* zopen-install/zopen-remove are modified to register installs and removals using the functions provided in analytics.sh
* The analytics are stored both locally (for user inspection) and sent to the remote server http://163.74.88.212:3000
* This PR also modifies the configuration logic to create a json file called config.json

Note:
* All of the statistics collected are anonymous, we use a UUID to denote a specific filesystem. 

TODOs:
* [x] Add documentation describing how collecting usage statistics work in zopen
* [x] Turn on via beta flag only for now ZOPEN_BETA_FEATURES=1
* [x] Set ZOPEN_IS_BOT in groovy scripts
* [x] Add tests: https://github.com/ZOSOpenTools/metaport/pull/44
* [x] allow analytics.json to be refreshed
* [x] Process historical install using logs
* [x] fix zopen remove statistics

Server TODOs:
* [x] Finish writing server code, available at http://163.74.88.212:3000
  * [x] Add charts, usage by uuid and usage by package links


bin/zopen-init
```diff
@@ -222,10 +214,63 @@ init()
...
+{
+  if isIBMHostname; then
+    isHostIBM=true
+  fi
+  if [ -n "$enableStats" ] && $enableStats; then
+    isCollectingStats=true
+  elif [ -n "$enableStats" ] && ! $enableStats; then
+    isCollectingStats=false
+  elif $isHostIBM; then
+    printAttention "Attention: Turning on usage statistics as your host is an IBM host. If this is incorrect, rerun with --noenable-stats"
```
MikeFultonDev, line 229: do we want ibmers to be able to override with ! $enableStats ?

bin/zopen-init
```diff
@@ -222,10 +214,63 @@ init()
...
+    isCollectingStats=true
+  else
+    printInfo "Do you want to contribute anonymous usage statistics to help improve z/OS Open Tools? (press y to collect stats or n for no): "
+    read collectStats < /dev/tty
+    if [ "y" = "${collectStats}" ]; then
```
MikeFultonDev, line 234: check for capital Y too and case-insensitive 'yes'

bin/zopen-init
```diff
@@ -222,10 +214,63 @@ init()
...
+      isCollectingStats=true
+    fi
+  fi
+
+  # Undocumented envar, for CI/CD tracking
+  if [ -n "${ZOPEN_IS_BOT}" ]; then
+    isBot=true
```
MikeFultonDev, line 241: any reason not to document it?

bin/zopen-init
```diff
@@ -222,10 +214,63 @@ init()
...
+  elif ${refresh}; then
+    printHeader "Refreshing zopen framework"
+  else
+    printHeader "Initializing zopen framework"
+  fi
+
+
+  printDebug "Validating input parameters"
+  case "${fslayout}" in
+    "usrlclz") zopen_pkginstall="usr/local/zopen"    ;;
```
MikeFultonDev, line 258: i thought we were going to get rid of these choices? although that could be a separate PR

bin/zopen-init
```diff
@@ -335,7 +382,41 @@ setDefaults()
...
+    printVerbose "You have chosen to not collect statistics."
+    return;
+  fi
+  jsonConfig="${rootfs}/etc/zopen/analytics.json"
+  if [ -e "${jsonConfig}" ]; then
+    return;
+  fi
+  jsonConfigWorking="${jsonConfig}.working"
+
+  UUID=$(generateUUID)
```
MikeFultonDev, line 408: should generateUUID take the host name or something as a parameter so people know how it's being constructed?

include/analytics.sh
```diff
@@ -0,0 +1,160 @@
...
+# This is the server that is collecting zopen usage stats
+ZOPEN_STATS_URL="http://163.74.88.212:3000"
+
+isAnalyticsOn()
+{
+  jsonConfig="${ZOPEN_ROOTFS}/etc/zopen/config.json"
+  if [ ! -f ${jsonConfig} ]; then
+    printError "config.json file does not exist. This should not occur. Please report an issue"
+  fi
+  isCollecting=$(jq -r '.is_collecting_stats' $jsonConfig)
```
MikeFultonDev, line 12: do you want to check the return code here in case 'jq' fails or is not present?

include/analytics.sh
```diff
@@ -0,0 +1,160 @@
...
+  if [ "$isCollecting" = "true" ] || [ -z "${ZOPEN_ANALYTICS_JSON}" ]; then
+    return 0
+  else
+    return 1
+  fi
+}
+
+isIBMHostname()
+{
+  ip_address=$(/bin/dig +short "$(hostname)" | tail -1)
```
MikeFultonDev, line 22: is /bin/dig in z/OS 2.4 and up? If so, all good.

include/analytics.sh
```diff
@@ -0,0 +1,160 @@
...
+    return 0
+  else
+    return 1
+  fi
+}
+
+sendStatsToRemote()
+{
+  json="$1"
+  response=$(curl -X POST -H "Content-Type: application/json" -d "$json" "${ZOPEN_STATS_URL}/statistics")
```
MikeFultonDev, line 34: how do you know curl is installed at this point?

include/analytics.sh
```diff
@@ -0,0 +1,160 @@
...
+
+  if ! isAnalyticsOn; then
+    return;
+  fi
+
+  # Local analytics
+  cat "${ZOPEN_ANALYTICS_JSON}" | jq '.removes += [{"name": "'$name'", "version": "'$version'", "timestamp": "'$(date +%s)'"}]' > "${ZOPEN_ANALYTICS_JSON}.tmp"
+  mv "${ZOPEN_ANALYTICS_JSON}.tmp" "${ZOPEN_ANALYTICS_JSON}"
+
+  uuid=$(jq -r '.profile' ${ZOPEN_ANALYTICS_JSON})
```
MikeFultonDev, line 110: can you add a comment - i don't know what it means to generate a uuid from the result of jq

include/common.sh
```diff
@@ -1143,4 +1151,45 @@ checkWritable()
   fi
 }
 
+generateUUID() 
+{
+  date_part=$(date +%s)
+  random_part=$((RANDOM))
+  uuid="$date_part-$random_part"
+  echo $uuid
```
MikeFultonDev, line 1159: how does this uuid relate to the one from jq above? Is the uuid from above this value that gets stored into the json?

MikeFultonDev (commented):
Questions more than any 'issues'

bin/zopen-init
IgorTodorovskiIBM: Is there a possibility that a third-party may get access to an ibm.com host and would not like there stats collected?

bin/zopen-init
IgorTodorovskiIBM: ok, I will turn this into a function

bin/zopen-init
IgorTodorovskiIBM: I wanted it to be exclusively by our CI/CD setup. If we expose it and document it, there could be a chance that can be used outside of our CI/CD.

bin/zopen-init
IgorTodorovskiIBM: Yeah, I think we should handle this in another PR, I just moved the code around in this change.

include/analytics.sh
IgorTodorovskiIBM: yep, `/bin/dig` is present in all the machines I've tested so far (2.4 and up)

include/analytics.sh
IgorTodorovskiIBM: I'll add a check here

bin/zopen-init
IgorTodorovskiIBM: It's currently defined in https://github.com/ZOSOpenTools/meta/pull/607/files#diff-8953cba62436c52982340b49f55152195f0a6119ee3fab71fba4e8f7d4049a16R1154 using date and $RANDOM. We could include hostname but I wanted to keep it as anonymous as possible for now.

v1gnesh:
> [ ]  Set ZOPEN_IS_BOT in groovy scripts

Say hello to multi-gitter :stuck_out_tongue: 
Can create a multi-gitterport repo for it now.

include/analytics.sh
```diff
@@ -176,3 +178,36 @@ registerError()
...
+
+processAnalyticsFromLogFile()
+{
+  log_file="$ZOPEN_LOG_PATH/audit.log"
+  if [ ! -f "${log_file}" ]; then
+    return
+  fi
+  encountered_packages=""
+
+  printHeader "Porcessing Analytics from log files for existing installation"
```
DevonianTeuchter, line 196: Porcessing ? ;-)

IgorTodorovskiIBM: 🐷essing oopsie

DevonianTeuchter:
@IgorTodorovskiIBM  After your demo yesterday, I was thinking... the plan is to write the anlalytic log in ${rootfs}/etc/zopen/analytics.json   - would it make more sense to add it to ${rootfs}/var/lib/zopen/analytics.json  ?  /etc is more for config, var/lib is for state data like this

IgorTodorovskiIBM:
> @IgorTodorovskiIBM After your demo yesterday, I was thinking... the plan is to write the anlalytic log in ${rootfs}/etc/zopen/analytics.json - would it make more sense to add it to ${rootfs}/var/lib/zopen/analytics.json ? /etc is more for config, var/lib is for state data like this

That makes sense, since we don't currently generate a var/lib/zopen directory, I'll need to also adjust --refresh to check if it exists and if not, create the dir

bin/zopen-init
MikeFultonDev: Hm. maybe. that's a good reason for people on an IBM machine to be able to override.

bin/zopen-init
MikeFultonDev: ok. it's exposed and not documented, but i am ok with that.

MikeFultonDev (approved):
LGTM

DevonianTeuchter (approved):
LGTM

PR #612, opened 2023-12-06
Docs: add a section for feedback

IgorTodorovskiIBM:


ijmitch:
Is email to @MikeFultonDev the only recommended feedback option? What about Discussions? (where you'd think it was happening already.)

lbdyck:
There should be a generic path for feedback and (imho) the Discussion area fits that bill.  Perhaps it needs to be renamed to Discussions/Feedback?

ijmitch:
@IgorTodorovskiIBM ah... then I do think you need to be more specific.

MikeFultonDev (approved):

PR #626, opened 2023-12-10
Metaport43 zopen upgrade failure

DevonianTeuchter:
Should fix ZOSOpenTools/metaport/#43
Fixes logic that populates the install/upgrade tree and causes issues with ports like git/githubcli, lua/luarocks, tree/zospstree ...


IgorTodorovskiIBM (approved):
nice! LGTM

Issue #613, opened 2023-12-06
Use better compression algorithm for pax files


AnthonyGiorgio:
The pax files containing the port releases are compressed using the `compress` algorithm (`.Z`).  This is an ancient, inefficient compression algorithm that has long been supplanted by better choices.  I suggest that we only use `.Z` for the minimal bootstrapping packages, and instead use `.xz` for everything else.  This will significantly reduce both download time and space on disk for the package cache.  It has the downside of making `xz` a dependency on the toolchain, but I think that's a reasonable tradeoff.

```sh
~ > ls -l dos2unix-master.20231114_165830.zos.pax.Z 
-rw-r-----   1 ANGIO    DBX       741888 Dec  6 08:18 dos2unix-master.20231114_165830.zos.pax.Z
~ > gunzip dos2unix-master.20231114_165830.zos.pax.Z 
~ > xz dos2unix-master.20231114_165830.zos.pax 
~ > ls -l dos2unix-master.20231114_165830.zos.pax.xz 
-rw-r-----   1 ANGIO    DBX       182528 Dec  6 08:18 dos2unix-master.20231114_165830.zos.pax.xz
```

IgorTodorovskiIBM:
Our pax.Z files are meant to work standalone as well. That's a great reduction though. We could provide both flavours in our releases, a .xz package which is consumed by zopen install (and users who have xz installed) and a pax.Z package for those who do not have either.

AnthonyGiorgio:
Traditionally, downloadable packages have been provided in multiple flavors.  Very early versions of GNU tools came in `.Z` and `.gz`.  When `bzip2` came out, they provided `.gz` and `.bz2`.  After `xz` was released. I saw all three versions supported for a bit.  Nowadays it seems that everything is mostly in `.xz` format.

IgorTodorovskiIBM:
This is really odd:
```
[ITODORO@ZOSCAN2B ~/projects]$ pax -w  -x pax -f meta.pax meta
[ITODORO@ZOSCAN2B ~/projects]$ pax -w -z -x pax -f meta.pax.Z meta
[ITODORO@ZOSCAN2B ~/projects]$ du -k meta.pax*
     77224 meta.pax
    101184 meta.pax.Z
     71936 meta.pax.zstd
```

How is meta.pax.Z larger in size than meta.pax?

Both xz and zstd (using -19 as the compression level) resulted in a 7.1mb file. xz was a few hundred bytes smaller.

[...]

PR #628, opened 2023-12-11
zopen-build: Generate a zstd-compressed asset, use sha256 checksum, change package naming convention

IgorTodorovskiIBM:
* Adds code to build a zstd compressed asset based on discussion in https://github.com/ZOSOpenTools/meta/issues/613#issuecomment-1849283271
* Changes metadata.json to specify an array of assets, now including pax.Z and .zst
* changes the package naming convention to ${PACKAGE_NAME}-${versionString}-${date}.zos.pax" so that we include the correct version string

### DIFF
--- a/bin/zopen-build
+++ b/bin/zopen-build
@@ -542,7 +542,7 @@ checkEnv()
     fi
   fi
 
-  implicitDeps="sed git jq curl"
+  implicitDeps="sed git jq curl xz zstd"
 
   if [ -z "${ZOPEN_COMP}" ]; then
     implicitDeps="${implicitDeps} comp_xlclang"
@@ -1507,7 +1507,7 @@ generateMetadataJSON()
 {
   printHeader "Creating Metadata JSON"
   package=${ZOPEN_NAME}
-  name=$(echo $package | cut -d "-" -f 1)
+  PACKAGE_NAME=$(echo $package | cut -d "-" -f 1)
   if [ -z "${versionString}" ]; then
     version=$(echo ${package} | cut -d "-" -f 2)
 
@@ -1537,7 +1537,7 @@ cat <<zz > "${ZOPEN_INSTALL_DIR}/metadata.json"
 {
 "version_scheme": "${metadata_version_scheme}",
 "product": {
-  "name": "${name}",
+  "name": "${PACKAGE_NAME}",
   "version": "${version}",
   "release": "${LOG_PFX}",
   "summary": "${name} on z/OS",
@@ -1545,6 +1545,7 @@ cat <<zz > "${ZOPEN_INSTALL_DIR}/metadata.json"
   "license": "${ZOPEN_REMOTE_URL}/blob/main/patches/LICENSE",
   "zopen_license": "${ZOPEN_REMOTE_URL}/blob/main/LICENSE",
   "categories": "${ZOPEN_CATEGORIES}",
+  "assets": [],
   "buildline": "${ZOPEN_BUILD_LINE}",
   "test_status": {
     "total_tests": "${totalTests}",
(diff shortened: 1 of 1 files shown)


v1gnesh:
The 'date' part can be found from the file's creation/modification time, right?
Typing this out to see how it looks :grin: 

**current:**
`zstd-1.5.5.20231124_101200.zos.pax.Z`

**linux/s390x:**
[`zstd-1.5.5-4.fc39.s390x.rpm`](https://rpmfind.net/linux/fedora-secondary/development/rawhide/Everything/s390x/os/Packages/z/zstd-1.5.5-4.fc39.s390x.rpm)

**zos/s390x:**
`zstd-1.5.5-4.zos24.s390x.pax.zst`

Makes me wonder... do we need to pax before running zstd on it?
pax _is_ our rough equivalent of rpm/deb... as a container of files.

The `-4` here refers to 4th release of the package at this `v.r.m`.

DevonianTeuchter:
By adding in the release, are we now publishing 3 versions (2.4/2.5/3.1) ??   Linux platforms add in the os release as they do publish different releases for those os's.  Eg. Fedora 39 has python3-productmd-1.37-1.fc39_1.38-1.fc39.noarch.drpm  and released similar for fc38.  The normal expectation for a user would be that if the package has a z/OS 2.4/2.5/3.1, then it will only apply to that version, not that it applies to that and later versions.  
Is s390x applicable to z/OS rather than Linux or is there a different naming convention (uname reports OS/390!!) - unless we are doing 24/31/64-bit versions then maybe those could be added.  
tbh, your original ${PACKAGE_NAME}-${versionString}-${date}&time.zos.pax made sense to me :-)

v1gnesh:
> are we now publishing 3 versions (2.4/2.5/3.1) ??

Setting up for it, hopefully. And dropping 2.4... soon.
We will end up setting up a "build matrix" in CI to do various combos.

I'm not entirely sure if, similarly, we need to keep separate builds for Go programs built in 1.19, 1.20, 1.21, etc.

> `python3-productmd-1.37-1.fc39_1.38-1.fc39.noarch.drpm`

In this case, it may be a deltarpm file, that contains just the stuff to go from 1.37 to 1.39 in fc39.


> The normal expectation for a user would be that if the package has a z/OS 2.4/2.5/3.1, then it will only apply to that version, not that it applies to that and later versions.

Yup, a package for a particular OS.

GOOS-GOARCH is `zos/s390x` so s390x still applies for z/OS.
s390 being 31-bit arch and s390x being 64-bit.

24/31/64 - :scream: - this is indeed an additional term we may need to add, assuming [these](https://github.com/zosopentools/?q=_31&type=all&language=&sort=) refer to ports that run in 31-bit.

v1gnesh:
Last 2 sentences of previous comment... I wonder... does that mean 31-bit builds should be called `zstd-1.5.5-4.zos24.s390.pax.zst`.

IgorTodorovskiIBM:
For C/C++ we're still targeting V2R4 (https://github.com/ZOSOpenTools/comp_clangport/blob/main/buildenv#L84). 

> > are we now publishing 3 versions (2.4/2.5/3.1) ??
> 
> Setting up for it, hopefully. And dropping 2.4... soon. We will end up setting up a "build matrix" in CI to do various combos.
> 
> I'm not entirely sure if, similarly, we need to keep separate builds for Go programs built in 1.19, 1.20, 1.21, etc.
> 
> > `python3-productmd-1.37-1.fc39_1.38-1.fc39.noarch.drpm`
> 
> In this case, it may be a deltarpm file, that contains just the stuff to go from 1.37 to 1.39 in fc39.
> 
> > The normal expectation for a user would be that if the package has a z/OS 2.4/2.5/3.1, then it will only apply to that version, not that it applies to that and later versions.
> 
> Yup, a package for a particular OS.
> 
> GOOS-GOARCH is `zos/s390x` so s390x still applies for z/OS. s390 being 31-bit arch and s390x being 64-bit.
> 
> 24/31/64 - 😱 - this is indeed an additional term we may need to add, assuming [these](https://github.com/zosopentools/?q=_31&type=all&language=&sort=) refer to ports that run in 31-bit.

Right now we're targeting 2.4 in our C/C++ builds: https://github.com/ZOSOpenTools/comp_clangport/blob/main/buildenv#L84 . Newer OS levels for z/OS are backwards compatible. So for us, this means the min OS level is 2.4, but it should work for 2.5, 3.1 as well. Perhaps we can use just "zos" as the release for now until we decide to add builds for each version?

DevonianTeuchter (changes requested):
You're last comment indicated to use zos as the release - the code seems to still be using zos24:
osRelease="zos24" #TODO: change based on build target
paxFileName="${PACKAGE_NAME}-${versionString}-${osRelease}.s390x.pax"

Otherwise, LGTM

MikeFultonDev (approved):
LGTM

v1gnesh:
Does anything need to go in before this, @IgorTodorovskiIBM?
Would be nice to see standard file names in the releases section :)

IgorTodorovskiIBM:
> Does anything need to go in before this, @IgorTodorovskiIBM? Would be nice to see standard file names in the releases section :)

Don't think so, I'll need to the sync the branch and do further testing to make sure I'm not breaking anything

PR #632, opened 2023-12-12
Add 'zip' as a supported extension

MikeFultonDev:
Changed two parts of the code - one where it checked for valid extensions and one where it now runs unzip instead of tar to unzip a zip file, if that is what is downloaded.

This is required for a new port I have started - groovy - which has it's binary stable builds saved as .zip on artifactory.

### DIFF
--- a/bin/zopen-build
+++ b/bin/zopen-build
@@ -840,7 +840,7 @@ extractTarBall()
   tarballz="$1"
   dir="$2"
   noext=${tarballz%%.zip}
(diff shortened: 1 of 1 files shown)


MikeFultonDev:
I ran the metaport tests against this code and verified it worked 
(as an aside, I uncovered a bug in my code with the metaport test, so that was very useful!)

IgorTodorovskiIBM (approved):
LGTM

PR #635, opened 2023-12-13
metadata.json sizes should be in bytes, not 512-byte units

IgorTodorovskiIBM:
Noticed that the metadata.json sizes are different than in the release cache.

By default:
> du measures file space in 512-byte units.

This adjusts it to bytes.


DevonianTeuchter (approved):
Lgtm

PR #638, opened 2023-12-15
Ensure z/OS standard sed is used in common.sh

DevonianTeuchter:
Use documented [:space:} char class when trimming rather than \t

### DIFF
--- a/include/common.sh
+++ b/include/common.sh
@@ -30,7 +30,7 @@ addCleanupTrapCmd(){
   if [ -n "${script}" ]; then
     for trappedSignal in "EXIT" "INT" "TERM" "QUIT" "HUP"; do
       newtrapcmd=$(echo "${script}" | while read trapcmd; do
-	       sigcmd=$(echo "${trapcmd}" | sed "s/trap -- \"\(.*\)\" ${trappedSignal}.*/\1/")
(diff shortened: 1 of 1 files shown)


include/common.sh
```diff
@@ -199,7 +199,7 @@ deleteDuplicateEntries()
 {
   value="\$1"
   delim="\$2"
-  echo "\${value}\${delim}" | awk -v RS="\${delim}" '!(\$0 in a) {a[\$0]; printf("%s%s", col, \$0); col=RS; }' | sed "s/\${delim}$//"
+  echo "\${value}\${delim}" | awk -v RS="\${delim}" '!(\$0 in a) {a[\$0]; printf("%s%s", col, \$0); col=RS; }' | /bin/sed "s/\${delim}$//"
```
MikeFultonDev, line 202: should this be zossed ?

DevonianTeuchter: This line is part of the generated zopen-config file.  zopen-config doesn't pull in the common.sh script where the zossed wrapper around /bin/sed is defined. As such, unless we include that wrapper in zopen-config, hardcoding to /bin/sed seems the simplest solution.

MikeFultonDev (approved):
Lgtm

PR #640, opened 2023-12-15
Cleanup `zopen-install --help` output

AnthonyGiorgio:
The output of `zopen-install --help` needed some work.  It was rather slow to print out, due to the large amount of `echo` statements used.  I converted it to a single `cat` call, along with fixing some spelling, grammar, and argument issues.

### DIFF
--- a/bin/zopen
+++ b/bin/zopen
@@ -23,9 +23,9 @@ setupMyself
 
 printHelp(){
 cat << HELPDOC
-zopen is a utility for managing a z/OS Open Tools environment.
+${ME} is a utility for managing a z/OS Open Tools environment.
 
-Usage: zopen [COMMAND] [OPTION] [PARAMETERS]...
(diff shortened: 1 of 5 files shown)


DevonianTeuchter (commented):
Might be worth removing the -h, --h, -syntax, --v and -verbose flags from the case statement around line 550 to keep the helpdoc in sync (and those prams being unused and inconsistent across other scripts)

MikeFultonDev:
> Might be worth removing the -h, --h, -syntax, --v and -verbose flags from the case statement around line 550 to keep the helpdoc in sync (and those prams being unused and inconsistent across other scripts)

+1

MikeFultonDev (approved):
Thank You! LGTM

AnthonyGiorgio:
I cleaned up some of the syntax help printouts.  There's more work to be done, but this is a good start.

PR #642, opened 2023-12-17
Add copyright text

AnthonyGiorgio:
This PR adds a copyright line to each script file.    It also sorts the command line arguments alphabetically for some of the scripts.  Finally, it continues to speed up help text printing by replacing repeated `echo` calls with a single `cat`.

### DIFF
--- a/bin/zopen
+++ b/bin/zopen
@@ -43,7 +43,7 @@ Command:
   upgrade           upgrades existing z/OS Open Tools packages.
 
 Options:
-  -h,-?, --help     display this help and exit.
+  -h, --help, -?    display this help and exit.
   -v, --verbose     run in verbose mode.
 
 Examples:
--- a/bin/zopen
+++ b/bin/zopen
@@ -15,7 +15,7 @@ setupMyself()
   MYDIR="$(cd "$(dirname "$0")" > /dev/null 2>&1 && pwd -P)"
   INCDIR="${MYDIR}/../include"
   if ! [ -d "${INCDIR}" ] && ! [ -f "${INCDIR}/common.sh" ]; then
-    echo "Internal Error. Unable to find common.sh file to source" >&2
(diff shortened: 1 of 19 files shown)


MikeFultonDev (commented):
Can you elaborate on the value of a copyright for open source?

AnthonyGiorgio:
> Can you elaborate on the value of a copyright for open source?

IP law is complicated and varied in different countries.  In the US, the code is automatically copyrighted upon being written, so having a statement like that at least makes the authorship clear.

 Having a copyright statement is pretty standard for open source projects - take a look at this [example from less](https://github.com/gwsw/less/blob/8f4691901c98aef748e48837650f9c847d0236b5/main.c#L2).

IgorTodorovskiIBM:
Changes look good overall.

Should the copyright start from the first date of publication? 2022?

Also, I'm noticing that copyright notices are less common in newer open source projects. I found this article which attempts to explain why it's no longer as valuable: https://opensource.com/article/20/10/copyright-notices-open-source-software

AnthonyGiorgio:
Yes, the date should start with the first date of publication.  Perhaps we could maintain the date(s) in the `zopen-version` file only?

MikeFultonDev:
> Changes look good overall.
> 
> Should the copyright start from the first date of publication? 2022?
> 
> Also, I'm noticing that copyright notices are less common in newer open source projects. I found this article which attempts to explain why it's no longer as valuable: https://opensource.com/article/20/10/copyright-notices-open-source-software

Thanks for the link. I didn’t think there was much value here to start and this seems to reinforce it. 

unless anyone objects I would like to not have copyrights in the code the z/OS open tools community merges. I will clean up the generic copyright I put in before which I feel now was a mistake. 
we should not change any of the licenses even if those do have some copyright text - we need to leave those as js.

MikeFultonDev (approved):
LGTM. Thanks!

PR #645, opened 2023-12-29
Reference docs styling

ijmitch:
This is just improving the rendering and styling of the `zopen` reference content.

Using the Docsify iframe embedding facility restores the light/dark theme choice - so that if you land on the pages in dark mode you can at least switch to the nicer light rendering on the page you're on. (Update 30/12/23) However, [aa46bb4](https://github.com/ZOSOpenTools/meta/pull/645/commits/aa46bb4e0993d1fc3fc56787394077f47900c51d) fixes the background colour regardless of the theme choice - but it's still better to have the button do something than be broken.

I've also injected some CSS to make the font sans-serif to match the rest of the site.


ijmitch:
I've now overridden the background colour for the `iframe` so that it is legible in both light and dark modes.
image

and 
image

Also suppressing the largely useless set of links at the top to the sections from displaying.

IgorTodorovskiIBM:
Very nice! Thanks! 

We'll need to additionally update the Jenkins job Port-Update-Docs as these files are dynamically generated.

I moved the code to the meta repo instead of inlinining it within Jenkins https://github.com/ZOSOpenTools/meta/blob/main/cicd/docupdate.sh#L28. You may need to rebase or sync your branch with the main branch.

cicd/docupdate.sh
```diff
@@ -0,0 +1,48 @@
...
+export ZOPEN_ROOTFS="na" # To workaround sourcing zopen-config error
+mkdir -p "man/man1/"
+zopen-help2man "man/man1/" # Generate man pages
+
+cat <<EOF > docs/reference/zopen-reference.md
+# zopen reference documentation
+This page provides information about the zopen interface. Click on any of the zopen commands listed below to access the reference guide describing how to utilize that command.
+EOF
+
+# Generate html and markdown pages
```
ijmitch, line 22: @IgorTodorovskiIBM ... does this look OK?

IgorTodorovskiIBM: Yes, thank you!

IgorTodorovskiIBM (approved):
LGTM

PR #646, opened 2024-01-02
Update release cache script to leverage release metadata.json

IgorTodorovskiIBM:
* Rather than extracting and calculating the size of each pax.Z file (which takes a long time), it reads the metadata.json asset information instead
  ***Note:** I've decided to skip older releases that do not contain a metadata.json file. 
* This PR also generates an additional latest json with suffix: `_latest.json` which only captures the latest releases. This can optimize certain queries in zopen that only examine the latest releases.

### DIFF
--- a/tools/create_release_cache.py
+++ b/tools/create_release_cache.py
@@ -12,6 +12,7 @@ import datetime
 import requests
 import subprocess
 import shutil
+import urllib
 from collections import OrderedDict
 
 """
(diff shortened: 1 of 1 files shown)


DevonianTeuchter (approved):
LGTM - and provides a framework for adding other asset types

MikeFultonDev (approved):
LGTM

PR #649, opened 2024-01-03
Fix date to timestamp conversion

IgorTodorovskiIBM:
* Fixes the issue in processAnalyticsFromLogFile as reported by @AnthonyGiorgio by embedding date into utilities:
```
parse error: Invalid numeric literal at line 1, column 10
FSUM9364 Bad date conversion: 2023-10-06 10:42:29
Usage: date [-cu] [-d time] [+format]
```
* Also embeds sha256sum, which would be used to verify the checksum matches the metadata.json's checksum

**Note:**
Another approach would be to embed coreutils (which includes date, sha256sum and many other tools). Another benefit of this is that we don't have to rely ascii workarounds like asciiecho (https://github.com/ZOSOpenTools/meta/blob/main/bin/zopen-build#L40) and we also can replace [numcpus.rexx](https://github.com/ZOSOpenTools/meta/blob/main/bin/numcpus.rexx) with `nproc`. It will increase size by 14mbs though.


HarithaIBM (approved):

include/analytics.sh
```diff
@@ -193,13 +193,19 @@ processAnalyticsFromLogFile()
...
+
+  dateCmd="${INCDIR}/../utilities/date"
+  if [ ! -e "$dateCmd" ]; then
+    return
+  fi
   encountered_packages=""
 
   printHeader "Processing analytics from existing installation's log files"
   grep 'handlePackageInstall.*Installed' "$log_file" | while IFS= read -r line || [ -n "$line" ]; do
+    set -x
```
AnthonyGiorgio, line 205: Is this for debugging?

AnthonyGiorgio (changes requested):

include/analytics.sh
IgorTodorovskiIBM: yep, removed it, thanks

MikeFultonDev (approved):
This is fine but can you create another issue so that we can pick up fixes to date/sha256sum if they are required as part of the build process of meta?

IgorTodorovskiIBM:
> This is fine but can you create another issue so that we can pick up fixes to date/sha256sum if they are required as part of the build process of meta?

Will do

PR #653, opened 2024-01-04
Add details about Clang and Go compiler in porting pre-req doc

IgorTodorovskiIBM:
* I also added it back to the sidebar as it was not readily available


MikeFultonDev (approved):
Lgtm

PR #666, opened 2024-01-19
fix zopen-promote help so that it can be properly converted to man pages

### LINKED ISSUE #663: zopen-promote issues
1. `man zopen-promote` is garbled (via ssh):
image

Under OMVS this is generated `groff: fatal error: cannot load 'DESC' description file for device 'utf8' `

2. `zopen-promote --help` displays the entire help - should pipe thru `less` for easier paging /reading

3. after it is complete it really isn't - it keeps going:
image

But it is really checking things and building the `etc/zopen-config` - there should be a message along with the spinning /

4. Ideally would like to be able to select which packages to promote.  For example I may only want to promote `git` and `bash` along with any prereqs and definitely do not want to promote `zopen` :)

MikeFultonDev:
Fixes #663 
The issue was that the fancy header was being printed out before the help, and the header fancy characters were causing help2man grief.
Fix was to move the fancy header to after the help/version info was printed out. 

I ran 'zopen build' for metaport against this. It failed the same way as without my fix (looks like metaport is busted).

### DIFF
--- a/bin/zopen-promote
+++ b/bin/zopen-promote
@@ -21,7 +21,7 @@ setupMyself
 
 printHelp(){
 cat << HELPDOC
-"${ME}" is a utility for z/OS Open Tools to generate a clone of
+${ME} is a utility for z/OS Open Tools to generate a clone of
 an existing zopen environment. For example, a user can install
 to a test area, validate the behavior, and promote to a production area.
 
@@ -89,7 +89,6 @@ HELPDOC
 
 args=$*
 
-printHeader "Promoting zopen environment"
 
 verbose=false
 debug=false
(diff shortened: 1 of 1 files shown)


MikeFultonDev:
this now runs clean after pulling in a separate fix for metaport

IgorTodorovskiIBM (approved):
LGTM

lbdyck:
tyvm

peter-sylvester:
The examples in zopen-promote --help don't seem to match the usage: 
1: Order of OPTIONS and DESTINATION
2: Did my eye blackspot miss the destination? Is a DESTINATION   '--notknown' allowed?
3: OPTIONS -> [OPTION ...]

Have fun

PR #690, opened 2024-02-07
Address issues in usage statistics feature

IgorTodorovskiIBM:
* Fixes various bugs discovered when tested on a bare environment
* Also changes the .gitattributes to use UTF-8 encoding


DevonianTeuchter (approved):
LGTM.

Issue #688, opened 2024-02-06
sudo uid(0) setuid ?


lbdyck:
used `zopen` to install `sudo` and it would not work - reported 

`/sudo must be owned by uid 0 and have the setuid bit set`

The installer should be aware of this requirement - shouldn't it?

lbdyck:
Or should `zopen install` and `zopen upgrade` be run as superuser?

lbdyck:
ok - I looked at the install for `sudo` and it has this

```

IMPORTANT NOTE: Installation of sudo is NOT COMPLETE.
For details on sudo, see: https://www.sudo.ws/releases/stable/#1.9.13p3
To finish installing sudo, run the following commands with elevated privileges:
BIN_SUDO='cvtsudoers sudo sudoedit sudoreplay'
SBIN_SUDO='sudo_logsrvd sudo_sendlog visudo'
SUDO_INSTALL_LOCAL=/home/splbd/zopen/usr/local/zopen/sudo/sudo-1.9.13p3.20231110_104125.zos
cd $SUDO_INSTALL_LOCAL/bin
cp $BIN_SUDO /usr/bin/
cd $SUDO_INSTALL_LOCAL/sbin
cp $SBIN_SUDO /usr/sbin/
cd /usr/bin
chown 0:0 $BIN_SUDO
cd /usr/sbin
chmod u+s $SBIN_SUDO
Review the $SUDO_INSTALL_LOCAL/etc/sudoers file.
Use visudo to create your own /etc/sudoers file.
```

This was lost among the dozen plus install packages and only because I scrolled back - this information, and any others like this, need to be more prominently presented to the user.

IgorTodorovskiIBM:
> ok - I looked at the install for `sudo` and it has this
> 
> ```
> 
> IMPORTANT NOTE: Installation of sudo is NOT COMPLETE.
> For details on sudo, see: https://www.sudo.ws/releases/stable/#1.9.13p3
> To finish installing sudo, run the following commands with elevated privileges:
> BIN_SUDO='cvtsudoers sudo sudoedit sudoreplay'
> SBIN_SUDO='sudo_logsrvd sudo_sendlog visudo'
> SUDO_INSTALL_LOCAL=/home/splbd/zopen/usr/local/zopen/sudo/sudo-1.9.13p3.20231110_104125.zos
> cd $SUDO_INSTALL_LOCAL/bin
> cp $BIN_SUDO /usr/bin/
[...]

PR #693, opened 2024-02-08
Download metadata.json during zopen install + support for installation caveats

IgorTodorovskiIBM:
* Download metadata.json from zopen install 
* The caveats portion covers issue https://github.com/ZOSOpenTools/meta/issues/688


DevonianTeuchter (approved):
LGTM! 
All makes sense given the initial issue that was raised!

PR #718, opened 2024-03-03
Adjusting --all option

emanuelhaine:
The --all option should handle all the tasks below:

- cache
- dangling
- metadata
- unused

It is not only a good practice but this adjusment will also match the standard of other package manager, such as yum/dnf.

### DIFF
--- a/bin/zopen-clean
+++ b/bin/zopen-clean
@@ -29,17 +29,17 @@ from the system to save space and prevent clutter.
 Usage: ${ME} [OPTION] [PACKAGE]
 
 Options:
-  --all             apply cleanup command to all applicable packages.
   -c, --cache  [PACKAGE ...]
                     cleans the downloaded package cache; packages will be
                     re-downloaded if needed.
   -d, --dangling    removes dangling symlinks from the zopen file system in
                     case of issues during package maintenance.
-  -h, --help, -?    display this help and exit.
   -m, --metadata    cleans and refreshes the metadata for zopen.
   -u, --unused [PACKAGE ...]
                     remove versions of PACKAGEs that are available as
                     alternatives, leaving only the currently active version.
+  --all             does all the above
+  -h, --help, -?    display this help and exit.
   -v, --verbose     run in verbose mode.
   --version         print version.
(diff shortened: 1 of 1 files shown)


IgorTodorovskiIBM (approved):
LGTM, we may also want to add an option to denote all packages at some point (something like `--allpackages`)

emanuelhaine:
So, what if I add an exit 0 on this block 

```
  "--all")
    all=true
    unused=true
    dangling=true
    cache=true
    meta=true
    cleanUnused
    cleanDangling
    cleanPackageCache
    cleanMetadata
```

And adjust the code to acomodate the --allpackages?

DevonianTeuchter:
That'd work :-)  
Note that the reluctance to have a generic "run all cleanup commands" was that the "dangling" check is **really** slow as it has to analyse all symlinks in the zopen filesystem for existence; and since symlinks are heavily used by design, that's a lot of checks! A handful of packages it'd be bearable but with all packages (or even just some of the bigger packages like ncurses & perl) things grind to a halt.  Something to contemplate... :-)

emanuelhaine:
I got it. I noticed the dangling find is using -exec and test to validate the symlinks. If we had the -xtype option available on find, we could use something like that:

`find ./ -type l -xtype l | xargs rm`

And it would be all the logic.  This way, dangling would be faster.

But for now, I'll work on adding the exit 0 and the --allpackages option and we can iterate and improve from this.

DevonianTeuchter:
> I got it. I noticed the dangling find is using -exec and test to validate the symlinks. If we had the -xtype option available on find, we could use something like that:
> 
> `find ./ -type l -xtype l | xargs rm`
> 
> And it would be all the logic.  This way, dangling would be faster.
> 
> But for now, I'll work on adding the exit 0 and the --allpackages option and we can iterate and improve from this.
> 
> 
Yeah, welcome to zos versions of utils like find that are quite old in support so are missing useful options 🙂

IgorTodorovskiIBM:
I'll close this for now, please re-open if you plan to revisit it.

Issue #720, opened 2024-03-04
clarify once-only use of zopen_root in quick start guide


covener:
In the QSG, there is a single reference to a `zopen_root` variable.  

I think in this context it should just use some default or something explicitly established in the same section.
Overriding could be mentioned as an aside/sub-bullet.

PR #723, opened 2024-03-07
Update QuickStart.md to address issue 720

IgorTodorovskiIBM:
To address https://github.com/ZOSOpenTools/meta/issues/720


DevonianTeuchter (commented):
Is it worth adding to the Note that by sourcing (depending on which packages are installed and sequence in .profile of automatic) some zopen-provided tools might run instead of natively-provided tools and so might have different behaviour...?

IgorTodorovskiIBM:
> Is it worth adding to the Note that by sourcing (depending on which packages are installed and sequence in .profile of automatic) some zopen-provided tools might run instead of natively-provided tools and so might have different behaviour...?

I'll handle this comment under https://github.com/ZOSOpenTools/meta/pull/766 with a suggestion of how to source zopen-config

DevonianTeuchter (approved):

PR #729, opened 2024-03-13
Update FAQ.md

mfsysprog:
Adding example .curlrc setup


mfsysprog:
Hmm, this examples uses --insecure, I will have to look at importing the certificate first before I publish this example

PR #730, opened 2024-03-13
Added some info about curl

mfsysprog:
Perhaps someone can add some text about how curl is used in the airgapped zopen setup that was (or will be) introduced.


DevonianTeuchter (approved):
LGTM
Is it worth including a link to the curl homepage as there are other options like .netrc support that users might want for credentials...?

mfsysprog:
Yes, sure, any additional information is welcome.

IgorTodorovskiIBM (approved):
LGTM

PR #731, opened 2024-03-14
Update GpgOnZOS.md

HarithaIBM:
Removed the term image as it might lead to expectation that it is container image


DevonianTeuchter (approved):
Lgtm - makes sense to make it generic

IgorTodorovskiIBM (approved):
LGTM

PR #750, opened 2024-04-11
Add iden information to executables

IgorTodorovskiIBM:
z/OS provides a tool called `iden` which can be used to display identity information about an executable. This PR adds such information:
* vendor, commit, and date of build

Then you can use `iden` as follows:
```
iden /path/to/git
/path/to/git:
     Vendor:zOS_Open_Tools BuildRev:3c2a3fd 2024-04-10 16:50:00 EDT
```

### DIFF
--- a/bin/zopen-build
+++ b/bin/zopen-build
@@ -2059,9 +2059,6 @@ generateZOSLIBHooks()
   if command -V "${ZOPEN_APPEND_TO_ZOSLIB_ENV_CODE}" > /dev/null 2>&1; then
     printVerbose "Appending additional environment zoslib variables..."
     zoslib_env="${zoslib_env}$(${ZOPEN_APPEND_TO_ZOSLIB_ENV_CODE})"
-  else
-    # Nothing to add, return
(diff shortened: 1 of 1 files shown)


DevonianTeuchter (approved):
One minor Q: there was an internal discussion that the string might get optimized out if not used - is that not the case (for all compilers) or does it need an additional #pragma string (or equivalent) - couldn't see a resolution in the discussion...

Otherwise LGTM

IgorTodorovskiIBM:
> One minor Q: there was an internal discussion that the string might get optimized out if not used - is that not the case (for all compilers) or does it need an additional #pragma string (or equivalent) - couldn't see a resolution in the discussion...
> 
> Otherwise LGTM

For context, in the internal examples, the global variable was defined as a static global. However, since the variable was not used within the source file, the compiler optimized it out.

In this case, I am not using the static qualifier which means the variable will not be optimized out by the compiler because another object could reference it.

HarithaIBM (approved):

PR #762, opened 2024-05-21
Remove escaped quotes in config

MacMalainey:
The escaped quotes were causing bad output when generating the `zopen-config` file during `zopen init`.

This change removes them.


DevonianTeuchter:
The code change appears reasonable but any chance of an explanation or paste a text-dump showing the "bad output" ? Would also allow searching on that bad output in future for example

IgorTodorovskiIBM (approved):
LGTM, turns out updated versions of /bin/sh are not evaluating the `\"`, so we see ZOPEN_ROOTFS=\"/home/yourid\" in the zopen-config

PR #763, opened 2024-05-22
Fix for zopen-build when zoslib_env is empty

IgorTodorovskiIBM:
This resulted in error messages when doing a zopen build


DevonianTeuchter (approved):
Lgtm

PR #765, opened 2024-05-23
Script to generate json file of vulnerabilities in packages

KeplerBoyce:
Adds a python script that gets the commit hashes for all releases of packages and queries the osv.dev database for vulnerabilities. The script is run in the docupdate cicd job and writes the vulnerability information to docs/api/zopen_vulnerability.json.

TODOs:
- [ ] Enable SSL for HTTP requests
- [ ] Add script for `zopen audit` command to view vulnerability info of installed packages


v1gnesh (approved):
Awesome, nice to see aiohttp.

tools/create_cve_json.py
```diff
@@ -0,0 +1,125 @@
...
+if args.verbose:
+    print("Fetching data from:", url)
+
+# Use the OSV API: 
+async def fetch_cves(session, commit):
+    url = "https://api.osv.dev/v1/query"
+    data = {
+        "commit": commit
+    }
+    async with session.post(url, json=data, ssl=False) as response:
```
IgorTodorovskiIBM, line 40: Future: add a todo comment so that we look into enabling SSL. I see it's set to false here

PR #766, opened 2024-05-23
Add alternate path for tools that conflict with /bin tools + mechanism to add them to your PATH

IgorTodorovskiIBM:
We're making modifications to move tools conflicting with /bin tools like Gawk's awk to an altbin/ directory.

To add altbin to their PATH, users can set ZOPEN_TOOLSET_OVERRIDE=1 and then source zopen-config.

You can also do this:
```
. /home/itodoro/zopen/etc/zopen-config --override-zos-tools
```

* Tools that conflict with /bin tools
  - coreutils - https://github.com/ZOSOpenTools/coreutilsport/pull/72/files
  - Sed - https://github.com/ZOSOpenTools/sedport/pull/12
  - findutils - https://github.com/ZOSOpenTools/findutilsport/pull/20
  - diffutils - https://github.com/ZOSOpenTools/diffutilsport/pull/31
  - Grep - https://github.com/ZOSOpenTools/grepport/pull/17
  - Awk - https://github.com/ZOSOpenTools/gawkport/blob/main/buildenv
  - Make - https://github.com/ZOSOpenTools/makeport/pull/42/files
  - Openssh - https://github.com/ZOSOpenTools/opensshport/pull/11
  - gawk: https://github.com/ZOSOpenTools/gawkport/pull/21/files

### DIFF
--- a/bin/zopen-build
+++ b/bin/zopen-build
@@ -1456,7 +1456,7 @@ export ${projectName}_HOME="\${PWD}"
 zz
 
   if [ -d "${ZOPEN_INSTALL_DIR}/bin" ]; then
-    echo "PATH=\"\${${projectName}_HOME}/bin:\$PATH\"" >> "${ZOPEN_INSTALL_DIR}/.env"
+    echo "PATH=\"\${${projectName}_HOME}/bin:\${${projectName}_HOME}/altbin:\$PATH\"" >> "${ZOPEN_INSTALL_DIR}/.env"
     echo "export PATH=\"\$(deleteDuplicateEntries \"\$PATH\" \":\")\"" >> "${ZOPEN_INSTALL_DIR}/.env"
   fi
   if [ -d "${ZOPEN_INSTALL_DIR}/lib" ]; then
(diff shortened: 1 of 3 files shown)


DevonianTeuchter (approved):
LGTM

lbdyck:
won't this lead to a violation of the principle of least astonishment with accompanying confusion?

IgorTodorovskiIBM:
> won't this lead to a violation of the principle of least astonishment with accompanying confusion?

I've added a caveat for tools that have an altbin. 

We can invert the defaults if preferred, such that conflicting tools are added to your PATH unless you explicitly say you dont want that.

lbdyck:
If I install a tool then I expect to use it and have it available. This should be up to the individual site/installer to decide how to handle things and it needs to be obvious. :)

IgorTodorovskiIBM:
> If I install a tool then I expect to use it and have it available. This should be up to the individual site/installer to decide how to handle things and it needs to be obvious. :)

In the case of gawk, it will still be available as gawk, but not as awk. `awk` will be available under altbin/. 

Brew has a similar caveat here: https://formulae.brew.sh/formula/gawk
> GNU "awk" has been installed as "gawk".
If you need to use it as "awk", you can add a "gnubin" directory
to your PATH from your ~/.bashrc and/or ~/.zshrc like:
PATH="$HOMEBREW_PREFIX/opt/gawk/libexec/gnubin:$PATH"

HarithaIBM (approved):

PR #773, opened 2024-06-03
Enable SSL for requests in create_cve_json script

### LINKED ISSUE #771: Enable SSL for HTTP requests in vulnerability json script
Currently, the create_cve_json python script from #765 performs requests with SSL disabled, as aiohttp raises exceptions otherwise ("Network is unreachable"). We will need to look into why this happens and how to enable SSL for these requests.

KeplerBoyce:
Closes #771.

### DIFF
--- a/tools/create_cve_json.py
+++ b/tools/create_cve_json.py
@@ -1,7 +1,9 @@
 import argparse
 import json
+import certifi
 import aiohttp
 import asyncio
+import ssl
 import sys
 from cvsslib import cvss31, calculate_vector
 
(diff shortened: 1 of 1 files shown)


v1gnesh:
Should we setup separate CI tests for shell and python scripts that go into the tools folder?

IgorTodorovskiIBM:
> Should we setup separate CI tests for shell and python scripts that go into the tools folder?

That's a good idea, let's open an issue for that

IgorTodorovskiIBM (approved):
LGTM

PR #779, opened 2024-06-05
Zopen config helper capabilities

DevonianTeuchter:
Provides a "zopen config  --get/--set" capbility to allow for set/get of zopen runtime environment variables, held in the <zopenrootfs>/etc/zopen/config.json file.  This is a naive --set in that there is [currently] no key/parameter validation, allowing storage of any configuration data in config.json as a key/value pair.  If a key already exists, it will be updated on a --set.


IgorTodorovskiIBM (approved):
Great addition, LGTM!

PR #780, opened 2024-06-05
Export environment key-value pairs

DevonianTeuchter:
Add the capability that when sourcing zopen-config, specifying the --knv or --eknv parameters will also output a key=value listing of the environment variables that zopen-config has calculated form the various appenv files.
This output can be captured to a file and used to source directly, noting that if packages change their appenv files in the future or if packages are added/removed, the static file will become stale and need to be regenerated by sourcing zopen-config again.

### DIFF
--- a/include/common.sh
+++ b/include/common.sh
@@ -166,20 +166,30 @@ writeConfigFile(){
   certPath="$4"
 
   cat << EOF >  "${configFile}"
+#!/bin/false  # Script currently intended to be sourced, not run
 # z/OS Open Tools Configuration file
 # Main root location for the zopen installation; can be changed if the
 # underlying root location is copied/moved elsewhere as locations are
 # relative to this envvar value
+displayHelp() {
+echo "usage: . zopen-config [--eknv] [--knv] [-?|--help]"
+echo "  --knv             Display zopen environment variables "
+echo "  --eknv            Display zopen environment variables, prefixed with an"
+echo "                   'export ' keyword for use in scripts"
+echo "  -?, --help            Display this help"
+}
+
 knv=false
 exportknv=""
 if [ \$# -gt 0 ]; then
   case "\$1" in
-    eknv) exportknv="export "; knv=true;;
-    knv) knv=true;;
+    --eknv) exportknv="export "; knv=true;;
+    --knv) knv=true;;
+    -?|--help) displayHelp; return 0;;
     *) echo "Error. Unknown parameter '\$1' passed to zopen-config." >&2; return 8;;
   esac
 fi
-[ \${knv} ] && env | sort > /tmp/zopen-config-env-orig.\$\$
+[ \${knv} ] && /bin/env | /bin/sort > /tmp/zopen-config-env-orig.\$\$
 
 ZOPEN_ROOTFS=\"${rootfs}\"
 export ZOPEN_ROOTFS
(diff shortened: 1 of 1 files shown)


DevonianTeuchter:
> any chance that the generated file can be compared to the `new` environment after a `zopen` upgrade/install/remove is performed and let the user know that the static file has changed and thus they may need to take some action?

The problem is, there is no (current) way to track where that static file is to do the comparison with, as that is down to the user.  Once (if) this PR ships and the re-worked installer does, it'd likely be possible to get a scriptlet to do this, exporting to a dedicated file somewhere (possibly configurable via a zopen-config setting).  
I guess the question is, does the functionality provided in this PR meet your needs (to write key=vals to screen to allow redirecting to a file), even if currently requires a wee bit of manual intervention?

lbdyck:
it works for me - even with the manual efforts required :)

THANK YOU

include/common.sh
```diff
@@ -244,6 +255,26 @@ LIBPATH=\${ZOPEN_ROOTFS}/usr/local/lib:\${ZOPEN_ROOTFS}/usr/lib:\$(sanitizeEnvVa
 export LIBPATH=\$(deleteDuplicateEntries "\${LIBPATH}" ":")
 MANPATH=\${ZOPEN_ROOTFS}/usr/local/share/man:\${ZOPEN_ROOTFS}/usr/local/share/man/\%L:\${ZOPEN_ROOTFS}/usr/share/man:\${ZOPEN_ROOTFS}/usr/share/man/\%L:\$(sanitizeEnvVar "\${MANPATH}" ":" "^\${ZOPEN_PKGINSTALL}/.*\$")
 export MANPATH=\$(deleteDuplicateEntries "\${MANPATH}" ":")
+
+if \${knv}; then
+  env | sort > /tmp/zopen-config-env-modded.\$\$
+  diffout=\$(diff /tmp/zopen-config-env-orig.$\\$ /tmp/zopen-config-env-modded.\$\$ | grep -E '^[>]' | cut -c3- )
```
IgorTodorovskiIBM, line 261: was this tested with both /bin/diff and z/OS Open Tools diff? Same question for grep. 
Wondering if we should just hardcode /bin/

include/common.sh
```diff
@@ -170,7 +170,18 @@ writeConfigFile(){
 # Main root location for the zopen installation; can be changed if the
 # underlying root location is copied/moved elsewhere as locations are
 # relative to this envvar value
-ZOPEN_ROOTFS="${rootfs}"
+knv=false
+exportknv=""
+if [ \$# -gt 0 ]; then
```
IgorTodorovskiIBM, line 175: Wondering if we need a --help to list out the supported options for zopen-config

include/common.sh
DevonianTeuchter: we should just hardcode /bin/diff to ensure consistency.

include/common.sh
DevonianTeuchter: > Wondering if we need a --help to list out the supported options for zopen-config
If we plan to add more capability to this script then it probably makes sense to, and perhaps to prefix them with '--' for consistency with other scripts.

DevonianTeuchter: Note, as this script is designed to be sourced into a users environment NOT run directly, adding in a blocking hashbang.

IgorTodorovskiIBM: I wonder if it makes sense to allow it be run in the case of --knv or --eknv since it's not altering the environment. But I think without the bash req, it's difficult to determine that so I'm ok with this.

IgorTodorovskiIBM:
Just to double check, was this tested under the /bin/sh shell?

IgorTodorovskiIBM (approved):
LGTM!

PR #783, opened 2024-06-07
Seperate docupdate.sh into two scripts, on_nightly and on_release

IgorTodorovskiIBM:
* Some of the scripts should run on a nightly basis, whereas others should run after a successful build of a tool
* This splits docupdate.sh into two scripts, on_nightly.sh and on_release.sh
* It also moves the logic for enabling workflows to on_nightly.sh, which avoids the 403 errors. FYI @v1gnesh


v1gnesh:
multi-gitter will expect the env `GITHUB_TOKEN` to be set in the CI system.
Assuming it's there, is it possible to check that it has succeeded in GH, or is this visible from the CI site only?

IgorTodorovskiIBM:
> multi-gitter will expect the env `GITHUB_TOKEN` to be set in the CI system. Assuming it's there, is it possible to check that it has succeeded in GH, or is this visible from the CI site only?

Yes, GITHUB_TOKEN is a credential managed and passed through our Jenkins CI

IgorTodorovskiIBM:
CI failing due to unrelated issue https://github.com/ZOSOpenTools/meta/issues/785

PR #811, opened 2024-06-13
Add quick install similar to brew

IgorTodorovskiIBM:
Similar to brew's one-liner: https://brew.sh/

Now that bash and curl are more accessible (provided via IBM Open Enterprise Foundation and Rocket Software), provide a one-liner to install zopen.

Screenshot 2024-06-13 at 1 26 56 PM

Since our intention is to deprecate/remove the [meta](https://github.com/ZOSOpenTools/meta/releases) release, the script downloads from metaport instead.

### DIFF
--- a/tools/zopen_install.sh
+++ b/tools/zopen_install.sh
@@ -1,0 +1,54 @@
+#!/bin/env bash
+# 
+# Quick Install tool for zopen's meta package
+# Assumes you have Curl and Bash installed (likely through Open Enterprise Foundation)
+
+if [[ $(uname) != "OS/390" ]]; then
+  echo "Error: This script is only for z/OS systems."
+  exit 1
+fi
+
+if ! command -v curl > /dev/null 2>&1; then
+  echo "Error: 'curl' command not found. Please install curl."
+  exit 1
+fi
+
+# Download the latest json
+url=$(curl --fail-with-body --silent -L https://raw.githubusercontent.com/ZOSOpenTools/meta/main/docs/api/zopen_releases_latest.json)
+if [ $? -gt 0 ]; then
+  echo "Error: Curl failed due to: \"$url\""
+fi
+
+# TODO: check if jq is present, and use it instead of this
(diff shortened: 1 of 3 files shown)


v1gnesh:
Maybe a bit late now... I wonder if this will be helpful for distribution & install - https://github.com/marwanhawari/stew

tools/zopen_install.sh
```diff
@@ -0,0 +1,63 @@
...
+
+# TODO: check if jq is present, and use it instead of this
+url=$(echo "$url" | /bin/tr ' ' '\n' |  grep "https://github.com/ZOSOpenTools/metaport/releases/download/" |  /bin/sed -e 's/.*\(https:\/\/github.com\/ZOSOpenTools\/metaport\/releases\/download\/[^"]*\.pax\.Z\).*/\1/')
+paxFile=$(basename "$url");
+
+echo "> Downloading z/OS Open Tools $url..."
+
+url=$(curl --fail-with-body -O -L "$url")
+if [ $? -gt 0 ]; then
+  echo "Error: Curl failed due to: \"$url\""
```
MikeFultonDev, line 32: tweak the message a bit? Is it that curl failed to download the zopen package manager from url $url ?

tools/zopen_install.sh
```diff
@@ -0,0 +1,63 @@
...
+if [ ! -f $paxFile ]; then
+  echo "Error: $paxFile not present"
+  exit 1
+fi
+
+# Extract meta.pax.Z
+echo "> Extracting $paxFile..."
+paxOutput=$(pax -rvf "$paxFile" 2>&1)
+if [ $? -gt 0 ]; then
+   echo "Error: Failed to unpax"
```
MikeFultonDev, line 45: should say what you were unpaxing...

IgorTodorovskiIBM:
> Maybe a bit late now... I wonder if this will be helpful for distribution & install - https://github.com/marwanhawari/stew

Interesting project, does a lot of similar things to zopen. If we decide to do a rewrite, perhaps we can leverage stew. 

I see it requires a GITHUB_TOKEN since it needs to access the github rest apis. That's one thing we tried to eliminate. I think we also want finer control on what tools to expose in zopen. This seems to work with all github projects. I'm a bit biased though ;) But the project has potential.

MikeFultonDev (approved):
thanks for incorporating my suggestions.

PR #828, opened 2024-07-09
Enable file system setting for how /bin tools are overriden via zopen-init

IgorTodorovskiIBM:
* Also adds an --override-zos-tools-subset option
  * Override a subset of /bin zos tools rather than all. Contains a subset of packages to override, delimited by newlines."

### DIFF
--- a/bin/zopen-init
+++ b/bin/zopen-init
@@ -430,8 +430,8 @@ generateJsonConfiguration()
 {
   "release_line": "${releaseLine}",
   "num_rm_procs": ${RM_FILEPROCS},
-  "is_collecting_stats": ${isCollectingStats}
-  "override_zos_tools": ${OVERRIDE_ZOS_TOOLS},
+  "is_collecting_stats": ${isCollectingStats},
+  "override_zos_tools": ${OVERRIDE_ZOS_TOOLS}
 }
 zz
     jq '.' "${jsonConfigWorking}" > "${jsonConfig}"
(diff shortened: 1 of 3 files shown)


DevonianTeuchter (commented):
The code in general looks good.
It is worth noting that given enough overrides and/or a deeply-nested zopen_rootfs, there could be an overflow of $PATH - probably unlikely but might be something to be aware of [hence adding this searchable comment] but not preventing merging.

Is it worth hard-coding the override file to a known location such as <rootfs>/etc/zopen/altbin-overrides for example - does making it configurable add anything?  [Perhaps allowing testing of different overrides via different filenames, but that's could done just by cp/mv'ing files in the /etc dir. instead]
Hardcoding would ensure consistency [all zopen config held in one location], can be documented as-such and can be requested if there are issues such as for servicability, just tarring <rootfs>/etc/zopen would get it all...

IgorTodorovskiIBM:
> The code in general looks good. It is worth noting that given enough overrides and/or a deeply-nested zopen_rootfs, there could be an overflow of $PATH - probably unlikely but might be something to be aware of [hence adding this searchable comment] but not preventing merging.
> 
> Is it worth hard-coding the override file to a known location such as /etc/zopen/altbin-overrides for example - does making it configurable add anything? [Perhaps allowing testing of different overrides via different filenames, but that's could done just by cp/mv'ing files in the /etc dir. instead] Hardcoding would ensure consistency [all zopen config held in one location], can be documented as-such and can be requested if there are issues such as for servicability, just tarring /etc/zopen would get it all...

The use case I was thinking of was 

> The code in general looks good. It is worth noting that given enough overrides and/or a deeply-nested zopen_rootfs, there could be an overflow of $PATH - probably unlikely but might be something to be aware of [hence adding this searchable comment] but not preventing merging.
> 
> Is it worth hard-coding the override file to a known location such as /etc/zopen/altbin-overrides for example - does making it configurable add anything? [Perhaps allowing testing of different overrides via different filenames, but that's could done just by cp/mv'ing files in the /etc dir. instead] Hardcoding would ensure consistency [all zopen config held in one location], can be documented as-such and can be requested if there are issues such as for servicability, just tarring /etc/zopen would get it all...

The scenario I was thinking about was if there's a shared read-only zopen file system and a user of the file system wants to enable zos overrides for coreutils and openssh, then they can use this option. We could have one on the zopen file system level as well in etc/zopen/

DevonianTeuchter:
I can see the value in that 😄
Just having the one override makes sense as otherwise it gets overly complex (precedence etc), so your implementation seems the sensible option.

DevonianTeuchter (approved):

PR #832, opened 2024-07-11
Add options to zopen-audit for dealing with vulnerable packages

### LINKED ISSUE #818: Add options in zopen audit to remove or upgrade packages with vulnerabilities
Add flags to `zopen audit` command to allow for removing all packages with vulnerabilities and upgrading all packages with vulnerabilities (if newer versions are available that resolve any vulnerabilities).

KeplerBoyce:
Closes #818.

Adds `--upgrade` and `--remove` flags to `zopen-audit` that offer two ways of handling vulnerable packages after scanning for vulnerabilities.

`zopen audit --upgrade` will check for newer releases for every package with vulnerabilities, report how many vulnerabilities will be resolved and added by upgrading, and give a y/n prompt for upgrading those packages:
```
CVE Summary:
5 vulnerabilities (2 low, 0 moderate, 1 high, 2 critical)

Newer releases are available for 2 packages with vulnerabilities.
Upgrading will resolve 2 vulnerabilities and add 1 vulnerability.

Packages: gpg  grafana

Would you like to upgrade these packages? [y/n]
```
`zopen audit --remove` will give a y/n prompt for removing all packages with vulnerabilities:
```
CVE Summary:
4 vulnerabilities (0 low, 1 moderate, 1 high, 2 critical)

There are 3 packages with vulnerabilities to remove.

Packages: git  grafana  openssh

***WARNING: Removing these packages may break dependencies of other packages.
Would you like to remove these pacakges? [y/n]
```

### DIFF
--- a/bin/zopen-audit
+++ b/bin/zopen-audit
@@ -26,8 +26,10 @@ in your installed packages.
 Usage: ${ME} [OPTION]
 
 Options:
-  -v, --verbose     run in verbose mode.
-  --version         print version.
+  -v, --verbose             run in verbose mode.
+  --version                 print version.
+  -u, --update, --upgrade   attempt to resolve vulnerabilities by upgrading packages.
+  -r, --remove              remove packages with vulnerabilities.
 
 Examples:
   zopen audit       check for vulnerabilities in all installed packages
@@ -41,6 +43,8 @@ HELPDOC
 args=$*
 verbose=false
 debug=false
+upgrade=false
+remove=false
 
 while [ $# -gt 0 ]; do
   printVerbose "Parsing option: $1"
@@ -56,6 +60,12 @@ while [ $# -gt 0 ]; do
   "-v" | "--verbose")
     verbose=true
     ;;
+  "-u" | "--update" | "--upgrade")
+    upgrade=true
+    ;;
+  "-r" | "--remove")
+    remove=true
+    ;;
   "--debug")
     # shellcheck disable=SC2034
     verbose=true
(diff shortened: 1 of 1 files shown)


bin/zopen-audit
```diff
@@ -81,16 +97,30 @@ downloadCVEJSONCache()
     printError "Failed to obtain vulnerability json from ${JSON_VULNERABILITIES_URL}; ${curlout}"
   fi
   chtag -tc 819 "${JSON_CVE_CACHE}"
+
+  if $upgrade; then
+    LATEST_RELEASES_CACHE="${cachedir}/zopen_releases_latest.json"
```
IgorTodorovskiIBM, line 102: Minor: Wonder if you can reuse or call downloadJSONCache() here: https://github.com/ZOSOpenTools/meta/blob/main/include/common.sh#L1167
It checks if the currently downloaded json is the latest and if not, then it downloads it
(this could be done in another PR)

KeplerBoyce: Maybe we could update zopen_vulnerability.json to include a timestamp as well, and add a json file name argument to the downloadJSONCache function so that we can use it for all three json caches (releases, latest releases, vulnerabilities)?

IgorTodorovskiIBM: We're tapping into the `ETag` header response here: https://github.com/ZOSOpenTools/meta/blob/b57de6ea09e1f6d1f548b412984b25ae03815aa4/include/common.sh#L1167

We could also extend that common function to download all 3

IgorTodorovskiIBM: We can handle this in another PR, will merge what you have

PR #833, opened 2024-07-11
Switch to new CVE json format

KeplerBoyce:
Currently, the `tools/create_vulnerability_doc.py` script errors when running the CI jobs:
```
  File "/var/lib/jenkins/workspace/Port-Update-Nightly/meta_update/tools/create_vulnerability_doc.py", line 87, in <module>
    for release, info in releases.items():
AttributeError: 'list' object has no attribute 'items'
```
This is because it assumes the new format of `docs/api/zopen_vulnerability.json` (the format was updated to make indexing by release timestamp more efficient), but this json file in the repo still has the old format. I thought that it would automatically update to the new format when `tools/create_cve_json.py` is run in the CI job, but the new format is never actually pushed to the repo because the CI job gets this error before it pushes the changes.

To fix this problem, we can manually update `docs/api/zopen_vulnerability.json` with this PR, and the CI jobs will no longer error and can return to automatically generating this json file.


IgorTodorovskiIBM (approved):
LGTM, thanks!

PR #834, opened 2024-07-13
Integrate AI with tinyllama model into zopen-build file

dianzhu:
Please review zopen-build file.

### DIFF
--- a/bin/zopen
+++ b/bin/zopen
@@ -45,6 +45,8 @@ Command:
 Options:
   -h, --help, -?    display this help and exit
   -v, --verbose     run in verbose mode
+  --ai              run with tinyllama model via LocalAI by using zopen build, zopen build
+                    will send error to the model and the model will provide what patch we need to do.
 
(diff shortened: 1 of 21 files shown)


bin/zopen
```diff
@@ -45,6 +45,8 @@ Command:
 Options:
   -h, --help, -?    display this help and exit
   -v, --verbose     run in verbose mode
+  --ai              run with tinyllama model via LocalAI by using zopen build, zopen build
```
IgorTodorovskiIBM, line 48: Can you just add --ai to zopen-build? The other commands shouldn't have this option at this point

dianzhu: > Can you just add --ai to zopen-build? The other commands shouldn't have this option at this point

I just updated it. Only zopen-build file has option --ai. The other commands won't have this option.

IgorTodorovskiIBM:
Closing this as it is now a stale PR.

PR #844, opened 2024-07-30
Add support for system prereq checks

IgorTodorovskiIBM:
Background: Now that we have projects targeting various OS levels and we're also planning to target environments that have /proc for certain tools - I think it's important to add pre-req checks on zopen-install.

-----
Developers can now add a system prereq check. To do this, they can add the following line to the buildenv:
```
ZOPEN_SYSTEM_PREREQS="procfs zos31"
```
This prereqs are added and propagated to the metadata.json. 

The zopen install script uses the metadata.json to determine what prereqs to check against.

The prereq scripts are maintained under meta's `include/prereq.sh`

Users can use zopen install --bypass-prereq-checks if they want to ignore the prereq checks.

The first project to use this will be procps, which will require the /proc filesystem and 2.5 or above.

Also bumps the metadata version number up


v1gnesh:
> Users can use zopen install --bypass-prereq-checks if they want to ignore the prereq checks.

If someone finds that they can "get away with" bypassing pre-req checks, and then discover that something is broken, we need a way to know that pre-req check(s) was bypassed, to speed up troubleshooting.

IgorTodorovskiIBM:
> > Users can use zopen install --bypass-prereq-checks if they want to ignore the prereq checks.
> 
> If someone finds that they can "get away with" bypassing pre-req checks, and then discover that something is broken, we need a way to know that pre-req check(s) was bypassed, to speed up troubleshooting.
> 
Good point!

DevonianTeuchter (approved):
LGTM

IgorTodorovskiIBM:
> In general, I like it and it looks good and useful 😁. My one design question would be the need for lots of scripts, one for each pre-req (and more as new prereqs are thought of) - is there a reason for this as opposed to being functions in a preq.sh included file for example (dynamic resolution/existence test of prereq to confirm support perhaps - though that can be done with functions and evals).

Good suggestion and that is now implemented in the latest revision

bin/zopen-install
```diff
@@ -398,6 +399,32 @@ handlePackageInstall()
       printf "${NC}${HEADERCOLOR}${BOLD}${name}${NC}:\n ${installCaveat}\n" >> ${caveatsFile}
     fi
 
+    
+    systemPrereqs=$(jq -r '.product.system_prereqs // empty | map(.name) | join(" ")' "${metadataFile}" 2>/dev/null)
+    if ! $bypassPrereqs; then
+      if [ -z "${systemPrereqs}" ]; then
+        systemPrereqs="${systemPrereqs} zos24" # set the min requirest as zos24
```
DevonianTeuchter, line 406: minor type in the comment

bin/zopen-install
```diff
@@ -398,6 +399,32 @@ handlePackageInstall()
...
+        else
+          . $ZOPEN_SYSTEM_PREREQ_SCRIPT
+          for prereq in $(echo "${systemPrereqs}" | xargs | tr ' ' '\n' | sort -u); do
+            printInfo "Checking system pre-req requirement $prereq"
+            if command -V "${prereq}" >/dev/null 2>&1; then
+              if ! ( ${prereq} ); then
+                printError "Failed system pre-req check \"$prereq\". If you wish to bypass this, install with --bypass-prereq-checks"
+              fi
+            else
+              printWarning "$prereq does not exist in $ZOPEN_SYSTEM_PREREQ_SCRIPT. You should upgrade meta. Bypassing prereq check."
```
DevonianTeuchter, line 420: We shouldn't bypass the prereq check if meta is in an odd state - either it's not updated enough to handle the specific prereq the package is trying to use or the package is broken and specifying an unknown prereq.

bin/zopen-install
```diff
@@ -398,6 +399,32 @@ handlePackageInstall()
...
 
+    
+    systemPrereqs=$(jq -r '.product.system_prereqs // empty | map(.name) | join(" ")' "${metadataFile}" 2>/dev/null)
+    if ! $bypassPrereqs; then
+      if [ -z "${systemPrereqs}" ]; then
+        systemPrereqs="${systemPrereqs} zos24" # set the min requirest as zos24
+      fi
+      if [ -n "$systemPrereqs" ]; then
+        if [ ! -f "$ZOPEN_SYSTEM_PREREQ_SCRIPT" ]; then
+          printWarning "$ZOPEN_SYSTEM_PREREQ_SCRIPT does not exist. You should upgrade meta. Bypassing prereq check."
```
DevonianTeuchter, line 410: would using -r to test for read-access to the script be a more robust check?  It would provide both a readability and existence check.  
Also, given that the code for doing the pre-req checks will only be from when this PR merges forward, is there a situation where meta is needing to be upgraded?  It would already be at the level that supports prereqs as any code to handle prereqs is not merged yet so the .product.system_prereqs  fields is ignored.  
The warning might be better as an error as the system is not setup correctly so we shouldn't install anything until it has been investigated.  This would require the message changing to something like  "$Z_S_P_S cannot be read. Check file permissions and reinstall the meta package or reinitialize the zopen environment if required. "  - or words to that effect?

DevonianTeuchter (changes requested):
In general, LGTM.  If there is an issue with meta or the prereq checking, we shouldn't bypass it with a warning - the issue should get resolved first otherwise a package could get installed when it perhaps shouldn't.

bin/zopen-install
IgorTodorovskiIBM: Makes sense, the warnings could lead to hidden bugs. Updated it

DevonianTeuchter (approved):
Lgtm

PR #852, opened 2024-08-22
GPG - sign & verify pax files

HarithaIBM:
`zopen-build `can now generate signed builds, and `zopen-install` can verify the signature before installation.

We can enable signed builds using the `--sign-pax` or `-sp` option. If this option is set, it reads the `ZOPEN_GPG_SECRET_KEY_FILE` and its passphrase from `ZOPEN_GPG_SECRET_KEY_PASSPHRASE_FILE` to sign the file.

These are mandatory for signing a file.

The public key is read from `ZOPEN_GPG_PUBLIC_KEY_FILE`. The public key and the signed file details are written to `metadata.json` (this JSON is not part of the PAX).

Since this is a major change, we have increased the `metadata.json` file version to 0.2.

Further, during installation, the signed file will be verified. We have a `--skip-verify` option that can be used if the `metadata.json` version is greater than`0.1` Otherwise, the verification of signed PAX files will take place.


bin/zopen-build
```diff
@@ -383,6 +383,7 @@ Option:
   --oci                 build and publish an OCI image to \$ZOPEN_IMAGE_REGISTRY.
   -s                    exec a shell before running configure. Useful when
                         manually building ports.
+  --signPax, -sp        This option signs the pax file.
```
IgorTodorovskiIBM, line 386: I noticed we use dashes in most cases to seperate words in our options. We should probably be consistent.
```
--sign-pax
```
@DevonianTeuchter , not sure if you have a convention already that we should follow?

bin/zopen-build
```diff
@@ -1832,13 +1883,42 @@ install()
...
+      fi
       setDepsEnv
 
       pax_size=$(getPathSizeInBytes "${paxFileName}")
       md5sum=$(md5sum "${paxFileName}" | awk '{print $1}')
-      jq '.product |= . + { "pax": "'${ZOPEN_NAME}.${LOG_PFX}.zos.pax.Z'", "pax_size": "'${pax_size}'", "md5": "'${md5sum}'" }' "${ZOPEN_ROOT}/install/metadata.json" > "${ZOPEN_ROOT}/install/metadata.json.tmp"
+
+      if [ -n "${signPax}" ]; then
+        if [ -z "${ZOPEN_GPG_SECRET_KEY_FILE}" ] || [ -z "${GPG_SECRET_KEY_PASSPHRASE}" ] || [ -z "${GPG_PUBLIC_KEY_PATH}" ]; then
+                printWarning "GPG keyring environment variables are not set!"
```
IgorTodorovskiIBM, line 1900: Can you consolidate this check with the one starting at line 1886? Perhaps use a boolean to indicate all the reqs are satisfied

bin/zopen-build
```diff
@@ -1832,13 +1883,42 @@ install()
 
       #TODO: Hack so that we can use coreutils md5sum without impacting builds
       ZOPEN_DEPS="${ZOPEN_DEPS} coreutils jq"
+      if [ -n "${signPax}" ]; then
+              if [ -z "${ZOPEN_GPG_SECRET_KEY_FILE}" ] || [ -z "${GPG_SECRET_KEY_PASSPHRASE}" ] || [ -z "${GPG_PUBLIC_KEY_PATH}" ]; then
```
IgorTodorovskiIBM, line 1887: Can you also check if GPG_PUBLIC_KEY_PATH and ZOPEN_GPG_SECRET_KEY_FILE are readable (-r)

bin/zopen-build
IgorTodorovskiIBM: Thinking about this more, if --sign-pax is specified, we should probably emit an error message. The user can just omit the --sign-pax option to bypass it.

bin/zopen-install
```diff
@@ -316,13 +376,11 @@ handlePackageInstall()
     printVerbose "Checking cache for already downloaded package [file name comparison]."
     location="zopen package cache"
   fi
-
   # Download the metadata json file
-  if ! runAndLog "curlCmd -L '${metadataJSONURL}' -o '${metadataFile}' ${redirectToDevNull}"; then
+  if ! runAndLog "curlCmd -L '${metadataJSONURL}' -o '${metadataFile}' "; then #${redirectToDevNull}"; then
```
IgorTodorovskiIBM, line 380: is this change intended?

bin/zopen-install
```diff
@@ -383,11 +441,25 @@ handlePackageInstall()
...
-      continue
+    extractPax=true
+    metadataVersion=$(jq -r '.version_scheme' "${metadataFile}" 2>/dev/null)
+    is_greater=$(echo "$metadataVersion > 0.1" | bc -l)
+
+    if [ "$is_greater" -eq 1 ] && [ "$skipverify" = "false" ]; then
+           verifySignatureOfPax
+           ret_val=$?
+           if [ $ret_val -eq 2 ]; then
+                   extractPax=false;
```
IgorTodorovskiIBM, line 452: So if the version is greater than 1, we enforce the signature verification, correct?
Should this print an error message in that case and exit?

IgorTodorovskiIBM: I noticed in verifySignatureOfPax, it can return 1 in certain error conditions. Should that case be handled here as well?

bin/zopen-install
```diff
@@ -21,6 +21,65 @@ setupMyself()
...
+
+  PUBLIC_KEY=$(jq -r '.product.public_key' "${metadataFile}")
+  if [ $? -ne 0 ] || [ "${PUBLIC_KEY}" = "null" ] || [ -z "${PUBLIC_KEY}" ]; then
+     printError "Failed to extract 'public_key' from ${metadataFile}" >&2
+     return 1
+  fi
+
+
+  # Create a temporary directory for GPG keyring
+  TMP_GPG_DIR="$HOME/tmp_gpg_verify"
```
IgorTodorovskiIBM, line 47: Might be better to reuse $tmp from here instead of $HOME: https://github.com/ZOSOpenTools/meta/blob/main/bin/zopen-build#L273

bin/zopen-install
```diff
@@ -21,6 +21,65 @@ setupMyself()
...
+  mkdir -p "$TMP_GPG_DIR"
+
+  SIGNATURE_FILE="signedfile.asc"
+  PUBLIC_KEY_FILE="scriptpubkey.asc"
+  printf "%b" "$SIGNATURE" | tr -d '"' | sed '/^$/d' > "$SIGNATURE_FILE"
+  printf "%b" "$PUBLIC_KEY" | tr -d '"' | sed '/^$/d' > "$PUBLIC_KEY_FILE"
+
```
IgorTodorovskiIBM, line 54: You should probably double check that the files exist since you re doing an rm -rf afterwards

bin/zopen-install
```diff
@@ -21,6 +21,65 @@ setupMyself()
...
+  printVerbose "Importing public key..."
+  gpg --no-default-keyring --keyring "$TMP_GPG_DIR/pubring.kbx" --batch --yes --import "$PUBLIC_KEY_FILE" > /dev/null 2>&1
```
IgorTodorovskiIBM, line 56: Check for errors here?

HarithaIBM:
Hi @IgorTodorovskiIBM Thank you for quick review. I have addressed all your comments.

bin/zopen-build
```diff
@@ -512,6 +519,52 @@ processOptions()
...
+signPaxFile()
+{
+  # Paths to your files
+  PRIVATE_KEY="${ZOPEN_GPG_SECRET_KEY_FILE}"
+  PASSPHRASE_FILE="${GPG_SECRET_KEY_PASSPHRASE}"
+  FILE_TO_SIGN="${paxFileName}"
+  OUTPUT_SIGNATURE="${paxFileName}.asc"
+
+  # Create a temporary directory for GPG keyring
+  TMP_GPG_DIR="$HOME/tmp_gpg"
```
IgorTodorovskiIBM, line 531: Can you use `$tmp` var from zopen-build here instead of $HOME?

bin/zopen-build
```diff
@@ -512,6 +519,52 @@ processOptions()
...
+  mkdir -p "$TMP_GPG_DIR"
+  chmod a+w "$TMP_GPG_DIR"
+
+  # Import the private key into the specified keyring files
+  printInfo "Importing private key..."
+  gpg --no-default-keyring --keyring "$TMP_GPG_DIR/pubring.kbx" --secret-keyring "$TMP_GPG_DIR/secring.gpg" --batch --yes --import "$PRIVATE_KEY" > /dev/null 2>&1
```
IgorTodorovskiIBM, line 537: Should be an error check here?

bin/zopen-build
```diff
@@ -512,6 +519,52 @@ processOptions()
...
+
+  # Verify that the key was imported successfully
+  printInfo "Checking if secret key is imported..."
+  gpg --no-default-keyring --keyring "$TMP_GPG_DIR/pubring.kbx" --secret-keyring "$TMP_GPG_DIR/secring.gpg" --list-secret-keys > /dev/null 2>&1
+  if [ $? -ne 0 ]; then
+      printWarning "No secret key found. Signing aborted."
+      rm -rf "$TMP_GPG_DIR"
```
IgorTodorovskiIBM, line 544: If you use $tmp from above, you could enhance cleanupOnExit to remove it on exit

bin/zopen-build
```diff
@@ -512,6 +519,52 @@ processOptions()
...
+
+  # Import the private key into the specified keyring files
+  printInfo "Importing private key..."
+  gpg --no-default-keyring --keyring "$TMP_GPG_DIR/pubring.kbx" --secret-keyring "$TMP_GPG_DIR/secring.gpg" --batch --yes --import "$PRIVATE_KEY" > /dev/null 2>&1
+
+  # Verify that the key was imported successfully
+  printInfo "Checking if secret key is imported..."
+  gpg --no-default-keyring --keyring "$TMP_GPG_DIR/pubring.kbx" --secret-keyring "$TMP_GPG_DIR/secring.gpg" --list-secret-keys > /dev/null 2>&1
+  if [ $? -ne 0 ]; then
+      printWarning "No secret key found. Signing aborted."
```
IgorTodorovskiIBM, line 543: Should this be a warning or error? In https://github.com/ZOSOpenTools/meta/pull/852/files#diff-1bce4db18a4653a30d1bb683ecb2a8b05317531e9167c6efbc09b357526a4d8eR554 you are using printError

bin/zopen-install
```diff
@@ -21,6 +21,76 @@ setupMyself()
...
+     return 1
+  fi
+
+  PUBLIC_KEY=$(jq -r '.product.public_key' "${metadataFile}")
+  if [ $? -ne 0 ] || [ "${PUBLIC_KEY}" = "null" ] || [ -z "${PUBLIC_KEY}" ]; then
+     printError "Failed to extract 'public_key' from ${metadataFile}" >&2
+     return 1
+  fi
+
+  if [ ! -d "${tmp}" ]; then
```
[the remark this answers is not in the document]
DevonianTeuchter: there is a function in common.sh - mktempdir - that uses the mktempfile function to wrapper that sort of $TMP checking and create unique dirs/files and also should register them with the cleanup handler...

include/common.sh
```diff
@@ -45,6 +45,39 @@ addCleanupTrapCmd(){
...
+  if [ ! -z ${tmp} ] && [ -d ${tmp} ]; then
+    break
+  fi
+done
+
+if [ ! -d "${tmp}" ]; then
+  printError "Temporary directory not found. Please specify \$TMPDIR, \$TMP or have a valid /tmp directory."
+fi
+
+# Capture start time before setting trap
```
IgorTodorovskiIBM, line 59: I think this timing part needs to be moved back to zopen-build

include/common.sh
```diff
@@ -45,6 +45,39 @@ addCleanupTrapCmd(){
   [ -n "${xtrc}" ] && set -x
 }
 
+# Temporary files
+for tmp in "${TMPDIR}" "${TMP}" /tmp; do
```
IgorTodorovskiIBM, line 49: Can you use a different var so that it doesnt collide with locals, maybe ZOPEN_TMP_DIR?

docs/Guides/ThePackageManager.md
```diff
@@ -144,3 +144,13 @@ The meta package, which include zopen, can be upgraded via the `zopen upgrade` c
...
+# Signing & verifying using zopen framework
+* `zopen-build` can now generate signed builds, and `zopen-install` can verify the signature before installation.
+
+* We can enable signed builds using the `--sign-pax` or `-sp` option. If this option is set, it reads the **ZOPEN\_GPG\_SECRET\_KEY\_FILE** and its passphrase from **ZOPEN\_GPG\_SECRET\_KEY\_PASSPHRASE_FILE** to sign the file. The public key is read from **ZOPEN_GPG\_PUBLIC\_KEY\_FILE**.
+<br/>These files are mandatory for signing a file.
+
+* The public key and the signed file details are written to metadata.json (this JSON is not part of the PAX).
+
+
+* Further, during installation, the signed file will be verified. We have a `--skip-verify` to bypass verification of signed PAX files.
```
IgorTodorovskiIBM, line 156:
```suggestion
* During installation, the signed file will be verified against the signature and public key. Not recommended:  If you prefer to bypass verification of signed PAX files, specify `--skip-verify`.
```

bin/zopen-install
```diff
@@ -383,11 +452,21 @@ handlePackageInstall()
...
-    if ! runLogProgress "pax -rf ${pax} -p p ${paxredirect} ${redirectToDevNull}" "Expanding ${pax}" "Expanded"; then
-      printWarning "Errors unpaxing, package directory state unknown."
-      printInfo "Use zopen alt to select previous version."
-      continue
+    extractPax=true
+    metadataVersion=$(jq -r '.version_scheme' "${metadataFile}" 2>/dev/null)
+    is_greater=$(echo "$metadataVersion > 0.1" | bc -l)
+
+    if [ "$is_greater" -eq 1 ] && [ "$skipverify" = "false" ]; then
+           verifySignatureOfPax
```
IgorTodorovskiIBM, line 460: Too much indentation here?

bin/zopen-install
```diff
@@ -383,11 +452,21 @@ handlePackageInstall()
...
     fi
+    if [ "$extractPax" = true ]; then
```
IgorTodorovskiIBM, line 462: Is extractPax needed anymore?

HarithaIBM: Yes it is set to false under certain conditions in verifySignatureOfPax

bin/zopen-install
```diff
@@ -21,6 +21,76 @@ setupMyself()
...
+           printInfo "${NC}${GREEN}Signature verified successfully: ${name}${NC}"
+           return 0
+        else
+           [ -e "${TMP_GPG_DIR}" ] && rm -rf "$TMP_GPG_DIR"
+	   extractPax=false;
+           printError "Verification failed"
+        fi
+   else
+        extractPax=false;
+        printError "Signature file does not exist."
```
IgorTodorovskiIBM, line 85: printError will exit the process meaning the subsequent check for extractPax is not needed.

bin/zopen-install
IgorTodorovskiIBM: Is there a case without the printError?

bin/zopen-install
```diff
@@ -383,11 +452,21 @@ handlePackageInstall()
...
 
-    if ! runLogProgress "pax -rf ${pax} -p p ${paxredirect} ${redirectToDevNull}" "Expanding ${pax}" "Expanded"; then
-      printWarning "Errors unpaxing, package directory state unknown."
-      printInfo "Use zopen alt to select previous version."
-      continue
+    extractPax=true
+    metadataVersion=$(jq -r '.version_scheme' "${metadataFile}" 2>/dev/null)
+    is_greater=$(echo "$metadataVersion > 0.1" | bc -l)
+
+    if [ "$is_greater" -eq 1 ] && [ "$skipverify" = "false" ]; then
```
IgorTodorovskiIBM, line 459:
```suggestion
    if [ "$is_greater" -eq 1 ] && ! $skipverify; then
```
This may be a bit more optimal

bin/zopen-install
```diff
@@ -21,6 +21,76 @@ setupMyself()
...
+  printVerbose "Checking if public key is imported..."
+  gpg --no-default-keyring --keyring "$TMP_GPG_DIR/pubring.kbx" --list-keys > /dev/null 2>&1
+  if [ $? -ne 0 ]; then
+      [ -e "${TMP_GPG_DIR}" ] && rm -rf "$TMP_GPG_DIR"
+      extractPax=false;
+      printError "No public key found. Verification aborted."
+  fi
+
+  # Verify the signature
+  printVerbose "Verifying the signature..."
```
IgorTodorovskiIBM, line 73:
```suggestion
  printInfo "Verifying the gpg signature..."
```

bin/zopen-install
```diff
@@ -21,6 +21,76 @@ setupMyself()
...
+  if [ -f "$SIGNATURE_FILE" ]; then
+        if gpg --no-default-keyring --keyring "$TMP_GPG_DIR/pubring.kbx" --verify "$SIGNATURE_FILE" "$FILE_TO_VERIFY" 2>&1 | grep -q "Good signature from"; then
+           printInfo "${NC}${GREEN}Signature verified successfully: ${name}${NC}"
+           return 0
+        else
+           [ -e "${TMP_GPG_DIR}" ] && rm -rf "$TMP_GPG_DIR"
+	   extractPax=false;
+           printError "Verification failed"
```
IgorTodorovskiIBM, line 81:
```suggestion
           printError "Verification failed."
```

bin/zopen-install
```diff
@@ -21,6 +21,76 @@ setupMyself()
 setupMyself
 checkWritable
 
+verifySignatureOfPax()
+{
+  # Extracting values and checking for errors
+  FILE_TO_VERIFY=$(jq -r '.product.pax' "${metadataFile}")
```
IgorTodorovskiIBM, line 27: You can add -e to the jq command and then just check the $? below:

```
       •   -e / --exit-status:

           Sets the exit status of jq to 0 if the  last  output  values  was  neither
           false  nor null
```

HarithaIBM:
Hi @IgorTodorovskiIBM @DevonianTeuchter @MikeFultonDev  - I have addressed the first round review comments. I have created a sample pipeline to generate signed build and i have also manually uploaded these signed builds to install & verify. Request your reviews to proceed on this further. Thank you.

bin/zopen-build
```diff
@@ -517,6 +492,53 @@ processOptions()
...
+  # Paths to your files
+  PRIVATE_KEY="${ZOPEN_GPG_SECRET_KEY_FILE}"
+  PASSPHRASE_FILE="${ZOPEN_GPG_SECRET_KEY_PASSPHRASE_FILE}"
+  FILE_TO_SIGN="${paxFileName}"
+  OUTPUT_SIGNATURE="${paxFileName}.asc"
+
+  # Create a temporary directory for GPG keyring
+  TMP_GPG_DIR="$zopen_tmp_dir/tmp_gpg"
+  mkdir -p "$TMP_GPG_DIR"
+  chmod a+w "$TMP_GPG_DIR"
```
IgorTodorovskiIBM, line 506: Doing a chmod a+w could be a security concern, suggest only adding +w to owner
```suggestion
  chmod u+w "$TMP_GPG_DIR"
```

bin/zopen-build
```diff
@@ -517,6 +492,53 @@ processOptions()
   done
 }
 
+signPaxFile()
+{
+  # Paths to your files
```
IgorTodorovskiIBM, line 497:
```suggestion
  # Paths to your files
  printHeader "Signing pax file ${paxFileName}"
```

bin/zopen-build
```diff
@@ -512,6 +519,52 @@ processOptions()
...
+  if [ $? -ne 0 ]; then
+      printWarning "No secret key found. Signing aborted."
+      rm -rf "$TMP_GPG_DIR"
+      return 1
+  fi
+
+  # Sign the file
+  printInfo "Signing the file..."
+
+  gpg --armor --no-default-keyring --keyring "$TMP_GPG_DIR/pubring.kbx" --secret-keyring "$TMP_GPG_DIR/secring.gpg" --batch --verbose --pinentry-mode loopback --passphrase-file "$PASSPHRASE_FILE" --default-key YOUR_KEY_ID --output "$OUTPUT_SIGNATURE" --detach-sig "$FILE_TO_SIGN" > /dev/null 2>&1
```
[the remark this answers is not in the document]
IgorTodorovskiIBM: @HarithaIBM - I think this still needs resolution

HarithaIBM:
@IgorTodorovskiIBM Thank you for the review. I have addressed all the comments. Please review.

bin/zopen-build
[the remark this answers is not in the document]
HarithaIBM: This is not needed anymore Igor, even --secret-keyring. So, i have removed these 2.

bin/zopen-build
```diff
@@ -512,6 +519,52 @@ processOptions()
...
+
+  if [ $? -ne 0 ]; then
+      printError "Error signing the file."
+      rm -rf "$TMP_GPG_DIR"
+      return 1
+  else
+      printInfo "File signed successfully. Signature saved to $OUTPUT_SIGNATURE."
+      SIGNATURE=$(jq -Rs . < ${OUTPUT_SIGNATURE})
+      echo "${SIGNATURE}"
```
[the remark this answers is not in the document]
HarithaIBM: I have removed this.

bin/zopen-install
HarithaIBM: I have modified this.

bin/zopen-install
IgorTodorovskiIBM: This is relatively minor, but would reduce the # of checks below - @HarithaIBM

IgorTodorovskiIBM (approved):
LGTM, one minor comment

IgorTodorovskiIBM (approved):
LGTM

PR #888, opened 2024-10-25
Mark initialization files as IBM-1047

IgorTodorovskiIBM:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Content Update

## Category

- [ ] zopen build framework
- [x] zopen package manager
- [ ] Documentation
- [ ] CI/CD
- [ ] Tools

## Description

The default codepage is IBM-1047, therefore we should provide all entrypoint files in IBM-1047. This would mean all env scripts, as they can then set AUTOCVT=ON.

## Related Issues

- Related Issue #
- Closes #

### DIFF
--- a/include/common.sh
+++ b/include/common.sh
@@ -1369,6 +1369,10 @@ promptYesOrNo() {
   return 0
 }
 
+# asciiecho: we are in the process of starting to use 'echo' from coreutils.
+# This echo will have a slightly different behaviour than the standard echo in that
+# the file will now be tagged as ascii instead of ebcdic
+# For compatibility, check the tag of the file after the echo and iconv the file IFF it is IBM-1047
 asciiecho()
 {
   text="$1"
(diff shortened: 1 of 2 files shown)


DevonianTeuchter (commented):
Can the comment about asciiecho be included in the moved code to explain what the function does/why it's needed?

include/common.sh
```diff
@@ -1369,6 +1369,38 @@ promptYesOrNo() {
   return 0
 }
 
+asciiecho()
```
IgorTodorovskiIBM, line 1372:
```suggestion
# asciiecho: we are in the process of starting to use 'echo' from coreutils.
# This echo will have a slightly different behaviour than the standard echo in that
# the file will now be tagged as ascii instead of ebcdic
# For compatibility, check the tag of the file after the echo and iconv the file IFF it is IBM-1047
asciiecho()
```

PR #908, opened 2024-11-14
Start gpg-agent for verification of signatures

HarithaIBM:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Content Update

## Category

- [ ] zopen build framework
- [x] zopen package manager
- [ ] Documentation
- [ ] CI/CD
- [ ] Tools

## Description
Start gpg-agent for verification of signatures if it is not already started

## Related Issues

- Related Issue #
- Closes #


bin/zopen-install
```diff
@@ -21,6 +21,25 @@ setupMyself()
 setupMyself
 checkWritable
 
+startGPGAgent()
+{
+  # Check if gpg-agent is running
+ if  ps -ef | grep -v grep | grep "gpg-agent" ; then #> /dev/null; then
```
IgorTodorovskiIBM, line 27: Can we avoid this check and just call `gpg-agent --daemon` since it already has logic to check if it's running?

```
gpg-agent --daemon
gpg-agent[67112018]: can't mmap pool of 32768 bytes: EDC5128I No such device. - using malloc
gpg-agent[67112018]: Please note that you don't have secure memory on this system
gpg-agent: a gpg-agent is already running - not starting a new one
```
note the last line

IgorTodorovskiIBM: ` gpg-agent --daemon; echo $?` gives a rc of 2 when it's already running and an rc of 0 when it has not been running

bin/zopen-install
```diff
@@ -65,7 +79,8 @@ verifySignatureOfPax()
   # Verify the signature
   printInfo "Verifying the gpg signature..."
   if [ -f "$SIGNATURE_FILE" ]; then
-        if gpg --no-default-keyring --keyring "$TMP_GPG_DIR/pubring.kbx" --verify "$SIGNATURE_FILE" "$FILE_TO_VERIFY" 2>&1 | grep -q "Good signature from"; then
+	gpg_output=$(gpg --no-default-keyring --keyring "$TMP_GPG_DIR/pubring.kbx" --verify "$SIGNATURE_FILE" "$FILE_TO_VERIFY" 2>&1)
```
IgorTodorovskiIBM, line 82: Should it be echo'ing gpg_output in the fail case?

bin/zopen-install
HarithaIBM: I have addressed to run daemon by default and once it is started , am checking to see if it is running or not.

bin/zopen-install
HarithaIBM: Right! I have displayed the output in failure case. Thank you.

IgorTodorovskiIBM (approved):
LGTM

PR #947, opened 2025-01-22
Modify startGPGAgent to test for existing socket

IgorTodorovskiIBM:
…rocess

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Content Update

## Category

- [ ] zopen build framework
- [x] zopen package manager
- [ ] Documentation
- [ ] CI/CD
- [ ] Tools

## Description

The `sleep 2` is unnecessarily slowing down zopen install. Test for the existence of sockets instead.

## Related Issues

- Related Issue #
- Closes #

### DIFF
--- a/bin/zopen-install
+++ b/bin/zopen-install
@@ -34,11 +34,11 @@ startGPGAgent() {
     if [ -r "$SOCKET_PATH" ]; then
       printVerbose "gpg-agent started successfully (socket created at $SOCKET_PATH)."
     else
-      printVerbose "gpg-agent started, but socket was not created at $SOCKET_PATH. Please verify your GPG installation."
+      printWarning "gpg-agent started, but socket was not created at $SOCKET_PATH. Please verify your GPG installation."
     fi
   else
     if [ -r "$SOCKET_PATH" ]; then
-      printVerbose "gpg-agent started successfully (socket created at $SOCKET_PATH), but gpg-agent returned a non-zero return code."
+      printWarning "gpg-agent started successfully (socket created at $SOCKET_PATH), but gpg-agent returned a non-zero return code."
     else
       printError "Failed to start gpg-agent. Reinstall or upgrade GPG using \"zopen install --reinstall gpg -y\" or \"zopen upgrade gpg -y\"."
     fi
--- a/bin/zopen-install
+++ b/bin/zopen-install
@@ -21,16 +21,28 @@ setupMyself()
 setupMyself
 checkWritable
 
-startGPGAgent()
-{
-     printInfo "- Starting gpg-agent..."
-     gpg-agent --daemon --disable-scdaemon
-     # Wait a moment to ensure the gpg-agent has time to start
-     sleep 2
-     # Check to confirm if gpg-agent started successfully
-     if ! ps -ef | grep -v grep | grep "gpg-agent" > /dev/null; then
-        printError "Failed to start the gpg-agent. Reinstall or upgrade GPG using \"zopen install --reinstall gpg -y\" or \"zopen upgrade gpg -y\" command."
-     fi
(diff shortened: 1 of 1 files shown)


DevonianTeuchter:
Is checking for socket existence the recommended way to check for an existing gpg-agent?
Working on my PR [that reworked some of the gpg logic], gpg-connect-agent seemed to work [with the /bye param], testing it's return code.  It can also start the agent, though the --no-scdaemon param would need to be in a .conf file [so a combination of gpg-connect-agent & gpg-agent commands perhaps].  Testing socket existence seems a bit of a workaround [much as ps output] but am happy to review as-is since it seems logically sound - just wondering if there is a cleaner/architected way!
If happy with the socket option, then lines 37 and 41 should probably be printWarning rather than info as gpg-agent hasn't worked as expected and might need investigating/configuration changes or monitoring incase it does not operate correctly.

IgorTodorovskiIBM:
> Is checking for socket existence the recommended way to check for an existing gpg-agent? Working on my PR [that reworked some of the gpg logic], gpg-connect-agent seemed to work [with the /bye param], testing it's return code. It can also start the agent, though the --no-scdaemon param would need to be in a .conf file [so a combination of gpg-connect-agent & gpg-agent commands perhaps]. Testing socket existence seems a bit of a workaround [much as ps output] but am happy to review as-is since it seems logically sound - just wondering if there is a cleaner/architected way! If happy with the socket option, then lines 37 and 41 should probably be printWarning rather than info as gpg-agent hasn't worked as expected and might need investigating/configuration changes or monitoring incase it does not operate correctly.

I came across  gpg-connect-agent /bye as well and yes, also noticed that it will start gpg agent automatically :)
When gpg was initially integrated in zopen, we didn't start gpg explicitly because we relied on gpg to automatically start it. However, there were some cases where if gpg is not configured correctly, it couldn't locate the gpg-agent executable. So we decided to start it by executing the gpg-agent from its known path.

But @HarithaIBM is working on a fix for gpg such that we don't have to rely on the config file to locate the gpg binaries, so once that goes in, we can probably remove this code.

bin/zopen-install
```diff
@@ -21,16 +21,28 @@ setupMyself()
...
+  if [ -r "$SOCKET_PATH" ]; then
+    printVerbose "gpg-agent is already running (socket found at $SOCKET_PATH)."
+    return 0
+  fi
+
+  if eval "$(gpg-agent --daemon --disable-scdaemon)" >/dev/null 2>&1; then
+    if [ -r "$SOCKET_PATH" ]; then
+      printVerbose "gpg-agent started successfully (socket created at $SOCKET_PATH)."
+    else
+      printVerbose "gpg-agent started, but socket was not created at $SOCKET_PATH. Please verify your GPG installation."
```
DevonianTeuchter, line 37: Should this be printWarning as users will not see the "verify" message unless running in verbose mode.

bin/zopen-install
```diff
@@ -21,16 +21,28 @@ setupMyself()
...
+    fi
+  else
+    if [ -r "$SOCKET_PATH" ]; then
+      printVerbose "gpg-agent started successfully (socket created at $SOCKET_PATH), but gpg-agent returned a non-zero return code."
```
DevonianTeuchter, line 41: Should this be printWarning as users will not see the non-zero rc to know there was an issue

DevonianTeuchter (changes requested):
General principle looks fine [as discussed].  Just a query on two message levels.

bin/zopen-install
IgorTodorovskiIBM: Updated to printWarning, thanks!

PR #976, opened 2025-03-07
Use shallow clones by default for build speedup

IgorTodorovskiIBM:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Content Update

## Category

- [x] zopen build framework
- [ ] zopen package manager
- [ ] Documentation
- [ ] CI/CD
- [ ] Tools

## Description

This should improve clone time by at least 10x. Most of the time we only need to a shallow clone, and the history is only important when debugging. The user can set ZOPEN_CLONE_FULL="yes" if they want to do a full clone. We could add an option if desired.

## Related Issues

- Related Issue #
- Closes #

### DIFF
--- a/bin/zopen-build
+++ b/bin/zopen-build
@@ -140,15 +140,18 @@ Restricted Usage (meta):
   ZOPEN_SHELL                     Specify an alternate shell to use if -s
                                   option specified. (defaults to /bin/sh)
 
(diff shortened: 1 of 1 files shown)


sachintu47 (approved):
LGTM

PR #985, opened 2025-03-14
Quiet zopen config

DevonianTeuchter:

- [ ] Refactor
- [ x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Content Update

## Category

- [ ] zopen build framework
- [ x] zopen package manager
- [ ] Documentation
- [ ] CI/CD
- [ ] Tools

## Description
Request from Discord to provide a quieter sourcing of zopen-config.
Added a --quiet flag that will suppress messages, other than errors or those that are expected (--knv and --eknv)

## Related Issues

- Related Issue #
- Closes #

### DIFF
--- a/include/common.sh
+++ b/include/common.sh
@@ -220,6 +220,7 @@ echo "      Override a subset of zos tools. Containing a subset of packages to o
 echo "  --knv                  Display zopen environment variables "
 echo "  --eknv                 Display zopen environment variables, prefixed with an"
 echo "                         'export ' keyword for use in scripts"
+echo "  --quiet                Do not display messages"
 echo "  -?, --help             Display this help"
 }
 EOF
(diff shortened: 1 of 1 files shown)


DevonianTeuchter:
> Great! Wondering if we should also suppress any output coming from?
> 
> https://github.com/zopencommunity/meta/blob/ac4d064204ef7560e9c1d6ddd912874353790456/include/common.sh#L319
> 
> ? Ideally there shouldn't be any, so I'm good with what you have.

I did initially think about masking the output from those - the issue is that if we redirect to /dev/null, we might lose errors so would need to effectively track error status from the individual scripts.  All doable but as you say, the scripts should not need to output anything so might be overkill.  We can always revisit in the future.

PR #1103, opened 2025-08-29
Issue 1089: Move the script generate_patches.sh from lipcap to meta tools

rshrotey-IBM:
Summary:
* Add a script to automatically generate patches to be committed to the port repos
* File finds the relevant commit-ids and generates the patches provided the source repo is setup correctly (please read instr- uctions)

Testing: Able to generate the patch files as expected

Limitation: Won't work in detached state

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Content Update

## Category

- [ ] zopen build framework
- [ ] zopen package manager
- [ ] Documentation
- [ ] CI/CD
- [ ] Tools

## Related Issues

- Related Issue #
- Closes #

### DIFF
--- a/tools/generate_patches.sh
+++ b/tools/generate_patches.sh
@@ -1,0 +1,137 @@
+#!/bin/bash
+
+# Script to generate the diff files/patches from source code repo to be committed to the port repo
+# Steps
+# 1. Apply all the patches and generate source code repo using zopen build command
+# 2. Commit the files modified by applying patches. Dev/bug fix changes should be done on top of this commit
+# 3. Once changes are ready, create new commit(s)
+# 4. Use this script to generate patch for files modified in step 3
+
+echo_error()
+{
+    echo "ERROR: $1"
+}
+
+get_commit_ids()
+{
+    BASE_COMMIT=      # source code gets cloned with this latest commit
+    PATCHED_COMMIT=   # Run zopen build to apply all the patches from port
+    CURRENT_COMMIT=   # Commit with further dev changes
+
+    #Check here if the branch is in detached state
+    if [ $(git branch -a | grep '* ' | grep -c "detached") != 0 ] 
+    then
+        echo_error "This is in detached state, please create a branch and try again!"
+        exit 1;
+    fi
+
+    #Find the current branch name
+    curr_branch=$(git branch -a | grep '* ' | cut -d ' ' -f 2)
+
+    #Find the commit from which curr_branch diverged
+    BASE_COMMIT=$(git reflog show ${curr_branch} | grep "branch: Created" | cut -d ' ' -f 1)
+
+    #Find the first commit after curr_branch branch diverged, this is the commit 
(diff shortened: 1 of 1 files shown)


sachintu47:
@rshrotey-IBM It didnt work for me. I got below errors

```
/home/sachint/test_gen_patch/meta>tools/generate_patches.sh -p /home/sachint/zostool_dev/dnf5port/ -s /home/sachint/zostool_dev/dnf5port/dnf5/
~/zostool_dev/dnf5port/dnf5 ~/test_gen_patch/meta
fatal: ambiguous argument '(no': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
Usage: grep [OPTION]... PATTERNS [FILE]...
Try 'grep --help' for more information.
fatal: ambiguous argument '(no': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
fatal: ambiguous argument '(no': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
ERROR: Something is wrong, BASE_COMMIT, PATCHED_COMMIT and CURRENT_COMMIT needed but are not set
```

Source code is checked out to a tagged version

tools/generate_patches.sh
```diff
@@ -0,0 +1,137 @@
+#!/bin/bash
```
IgorTodorovskiIBM, line 1:
```suggestion
#!/usr/bin/env bash
```
Bash may not always be under /bin

IgorTodorovskiIBM:
Looks good overall:
A couple of suggestions:
* A --dry-run option would be useful if the user wants to see what patches would be created without creating them.
* The default behaviour is to clobber existing files. Should there be a --no-clobber option in case users just want to create patches for new file changes.
* What happens if you have a file of the same name in different directories? src/foo and lib/foo. Will both the patches be created as patches/foo.patch?

tools/generate_patches.sh
rshrotey-IBM: Done

rshrotey-IBM:
> Looks good overall: A couple of suggestions:
> 
> * A --dry-run option would be useful if the user wants to see what patches would be created without creating them.
Done

> * The default behaviour is to clobber existing files. Should there be a --no-clobber option in case users just want to create patches for new file changes.
> I haven't yet encountered a situation where I wanted to generate patches for just the newly modified files. It's always been - regenerate the patches for all the modified files in a commit. But if this feature is needed I'd be happy to add it. 

> * What happens if you have a file of the same name in different directories? src/foo and lib/foo. Will both the patches be created as patches/foo.patch?
Yes, it will generate both the files, patches/src/foo.patch and patches/lib/foo.patch

PR #1126, opened 2025-11-10
Introduce :bin for dependencies to export only bin path

sachintu47:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Content Update

## Category

- [ ] zopen build framework
- [ ] zopen package manager
- [ ] Documentation
- [ ] CI/CD
- [ ] Tools

## Related Issues

- Related Issue #
- Closes #


IgorTodorovskiIBM:
Another consideration is that we could try to separate the .appenv into 2:
.appenv for runtime (and use the existing zopen_append_to_env) 
and create another hook zopen_append_to_library that will generate a .libraryenv file. 
If you add curl as a dep, it will source both .appenv and .libraryenv. If you add curl:bin as a dep, it will source .appenv and if you add curl:lib as a dependency, it will source .libraryenv

PR #1228, opened 2026-07-24
Add opt-in Python wheel publishing pipeline

IgorTodorovskiIBM:
## Summary

- generate `PUBLISH_PYTHON_WHEEL=true` for both dev and stable Python ports while keeping non-Python ports disabled
- stage wheels produced under `install/**`, archive them independently from pax/RPM artifacts, and route opted-in builds to a dedicated `Python-Publish` Jenkins job
- publish to the `wheels` Pulp Python repository and verify pure-Python wheels with pip against the public index
- make wheel retries idempotent by comparing exact filenames and SHA-256 hashes before upload, rejecting conflicting content, and rechecking after concurrent uploads
- make wheel metadata compatibility rewrites deterministic and RECORD-correct

## Validation

- `sh -n bin/zopen-generate bin/zopen-publish`
- syntax-checked the embedded Bash blocks in `cicd/publish_python.groovy`
- `git diff --check`
- generator smoke tests: Python stable and dev enabled; CMake stable and dev disabled
- mocked retry cases: identical retry skipped, conflicting retry rejected, malformed index rejected, absent wheel uploaded, and concurrent identical upload accepted
- deterministic compatibility rewrite test produced identical SHA-256 output across repeated runs
- live read-only retry check against the existing public test wheel correctly skipped upload

## Rollout notes

- create/configure a Jenkins `Python-Publish` job backed by `cicd/publish_python.groovy`
- provide Jenkins string credentials `PULP_USERNAME` and `PULP_PASSWORD`, or override their credential IDs through job parameters
- correct the Pulp `CONTENT_ORIGIN` public URL separately before production rollout; this is server configuration and is intentionally outside this repository change
- platform-specific wheel installation still requires validation on a compatible node; this pipeline verifies exact public-index presence for those wheels

### DIFF
--- a/bin/zopen-generate
+++ b/bin/zopen-generate
@@ -775,6 +775,10 @@ createCICD() {
   buildline=$1; touch "${name}port/cicd-${buildline}.groovy" && chtag -tc 819 "${name}port/cicd-${buildline}.groovy"
   buildlineUpper=$(echo "${buildline}" | awk '{print toupper($0)}')
   _groovy_description=$(echo "$description" | sed "s/'/\"/g")
+  publish_python_wheel=false
+  if [ "${build_system}" = "Python" ] && [ "${buildline}" = "stable" ]; then
+    publish_python_wheel=true
+  fi
   cat << EOT > "${name}port/cicd-${buildline}.groovy"
 node('linux') {
   stage ('Poll') {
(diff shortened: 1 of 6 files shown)


sachintu47 (approved):
LGTM

